PR #2, opened 2020-06-08
Add core module in Web SDK

jesusbotella:
☝️

### DIFF
--- a/src/lib/core/mixins/WithEvents.ts
+++ b/src/lib/core/mixins/WithEvents.ts
@@ -1,0 +1,49 @@
+import mitt from 'mitt';
+import { CartoError } from '../errors/CartoError';
+
+export abstract class WithEvents {
+  protected emitter = mitt();
+  protected availableEvents = ['*'];
+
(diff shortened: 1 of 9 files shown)


src/lib/core/mixins/WithEvents.ts
```diff
@@ -0,0 +1,49 @@
+import mitt from 'mitt';
+import { CartoError } from '../errors/CartoError';
+
+export abstract class WithEvents {
```
VictorVelarde, line 4: ouuu, I see what your're doing here Rick... 👏

VictorVelarde:
Is it ready now? It looks good, but I miss the tests...

And I was just wondering if we should include the namespaced exports on each migration PR

jesusbotella:
I have just added the missing tests from the original module. Everything seems to be working OK.

About the namespacing, I wouldn't export anything besides the credentials from core to the public, and probably we should move that to another folder eventually.

VictorVelarde (approved):
🆗  Take note of the next refactor steps, and let's move on with this

PR #7, opened 2020-06-09
Add Viz module from Toolkit

jesusbotella:
Tests are failing but cannot understand why. Seems like Jest/TS cannot import `mitt` in our tests. Gonna check if some configuration or routes changed.


VictorVelarde (approved):
💚  vamos!

PR #10, opened 2020-06-10
Add exported api

VictorVelarde:
Including basic stuff, coming from core / viz and data


VictorVelarde:
One comment for the review: I have moved `index.ts` to `src/lib`. It feels more natural to me and it allows also local imports like `from './core/Credentials'` (my intellisense on @ broke again).

I also tried to split the exports of the API as much as possible, to favour an 'easy read' over a less verbose style. We'll probably have some discussions or even refactors on that part, and that is what I was trying to ease.

jesusbotella (changes requested):
It looks good in overall, but I would avoid importing all modules from "child" modules in the main `index.ts` file.

I think we should just have "one line" per module, like:
```
import { style, Layer, Popup, basemap, sources } from './viz';
export { ...basemap, ...sources, ...style, Layer, Popup };
```

That way we would keep the modules kind of scoped within their own folders. What do you think?

VictorVelarde:
Thx @jesusbotella  for your comment, I agree with them, but I foresee changes in those pieces and how they combine, and that's why I wanted to lay out of all of them, to see its structure. 

Once we consolidate them, we can 'rejoin' them, but I would prefer to have the conversation with those pieces.

jesusbotella (approved):
As  @VictorVelarde pointed out, let's merge this, and we'll handle that later.

PR #11, opened 2020-06-11
Category Widget

jesusbotella:
This PR implements `CategoryWidget` and a base class for widgets with common methods.

## Category Widget's API
Category Widget does not have any public methods until now, so you just need to instantiate the class to initialize and populate the widgets with DataView's data.
```js
const widgetOptions = { valueFormatter: value => `${value}€`};

new carto.viz.CategoryWidget('as-category-widget', dataView, options);
```

### Pending
- [x] Export CategoryWidget in public API due to pending PR


package.json
```diff
@@ -72,11 +72,12 @@
     "webpack-cli": "^3.3.11"
   },
   "dependencies": {
-    "@deck.gl/core": "8.1.7",
-    "@deck.gl/geo-layers": "8.1.7",
-    "@deck.gl/google-maps": "8.1.7",
-    "@deck.gl/layers": "8.1.7",
-    "@deck.gl/mesh-layers": "8.1.7",
+    "@deck.gl/core": "8.1.9",
```
VictorVelarde, line 76: Nice, this one includes @juandjara contribution!!
Very soon we'll upgrade to alpha.3

src/lib/viz/index.ts
```diff
@@ -23,3 +23,6 @@ export {
 
 // Basemap helpers
 export { createGoogleMap, createMap } from './basemap';
+
+// Widgets
+export { CategoryWidget } from './widgets';
```
VictorVelarde, line 28: I think we can already start with a `carto.widget` namespace, for a CategoryWidget class there, as you proposed... what do you think?

src/lib/viz/widgets/widget.ts
```diff
@@ -0,0 +1,44 @@
...
+  constructor(element: string | HTMLElement, dataView: DataView) {
+    const domElement = queryDOMElement(element);
+    validateParameters(domElement, dataView);
+
+    this.element = domElement as HTMLElement;
+    this.dataView = dataView;
+  }
+
+  protected bindEvents() {
+    this.dataView.on('dataUpdate', () => this.updateData());
```
VictorVelarde, line 18: so we're assuming here that this is the only and relevant event for any dataview update, right? And this way, we have the 'widget update' logic for all of them....

/cc @manmorjim

src/lib/viz/widgets/widget.ts
```diff
@@ -0,0 +1,44 @@
+import { DataView } from '@/data/dataview/dataview';
+import { CartoError } from '@/core/errors/CartoError';
+import { queryDOMElement } from '../utils/dom';
+
+export class Widget {
```
VictorVelarde: Should this be an abstract class?

VictorVelarde (approved):
Just small suggestions. LGTM 🚀

src/lib/viz/widgets/widget.ts
jesusbotella: Yes, that's in fact what I thought. So it doesn't make sense to add a new event. We should rename this one to `dataChange` or merge both events to `dataUpdate` in the DataView.

src/lib/viz/index.ts
jesusbotella: I think we should move them when the namespace refactor happens :)

src/lib/viz/widgets/widget.ts
jesusbotella: Yes, gonna change it right now and merge

PR #13, opened 2020-06-11
Filters Collection

jesusbotella:


src/lib/index.ts
```diff
@@ -50,6 +54,7 @@ const styles = {
 export const viz = {
   ...sources,
   ...basemaps,
+  ...widgets,
```
VictorVelarde, line 57: As I said in the previous, and I think this namespace won't have much discussion --> let's go for an `export const widget = { ... widgets } `?

src/lib/data/dataview/dataview.ts
```diff
@@ -26,6 +27,14 @@ export class DataView extends WithEvents {
...
   }
 
+  addFilter(widgetId: string, filter: Filter) {
+    this.dataSource.addFilter(widgetId, { [this.column]: filter });
+  }
+
+  removeFilter(widgetId: string) {
+    this.dataSource.removeFilter(widgetId);
+  }
+
```
manmorjim: What about naming the parameter as `filterId` rather than `widgetId`? I think these filters will be used not just for widgets but other filters like polygons, circles, etc.

manmorjim (approved):

manmorjim:
It looks really great! With this approach I think I can override the methods `addFilter` and `removeFilter` of the class `DataView` in order to reinstantiate the map to get the filtered data using the Dataviews API.

 I've just added on comment about the naming of the parameters of those methods :blush:

VictorVelarde:
I'm having a look at conflicts

src/lib/data/dataview/dataview.ts
jesusbotella: Yes, it sounds good :)

jesusbotella:
@manmorjim Yes! You would need to override those methods in your source and use the FiltersCollection to group them all. And then, implement a SourceFilterApplicator or whatever you want to get those filters transformed into SQL.

@VictorVelarde Let me take a look at the conflicts :)

VictorVelarde (approved):
🚀 VAMOS!

PR #14, opened 2020-06-12
[Dataview] - Category (II) - Source mode

manmorjim:


src/lib/data/dataview/category/CategorySourceDataView.ts
```diff
@@ -0,0 +1,112 @@
...
+    }
+
+    const { aggregation, categories, count, max, min, nulls } = aggregationResponse;
+
+    const adaptedCategories = categories.map(({ category, value }) => {
+      return {
+        name: category,
+        value
+      };
+    });
```
VictorVelarde, line 85: I think we should just throw an error (and catch it in the client), instead of using an evented error approach... for these dataviews and in general for the whole lib. Aggree on this @jesusbotella ?

VictorVelarde (approved):
I made a comment regarding error strategy. As we want to land this soon, and then refactor with the 'new approach' ™️ I think we can apply those changes (and also file re-naming) in the next PR, with both local & server classes

PR #16, opened 2020-06-12
Refactor: Category DataViews with 3 modes (local / server: global + viewport)

manmorjim:
Discussion: https://www.notion.so/Discussion-about-dataviews-mode-342bc4d721404893bdcdaa13fc6a2c25

## Class Diagram

category_uml


src/lib/dataviews/DataViewWrapperBase.ts
```diff
@@ -0,0 +1,96 @@
...
+    this.dataviewImpl.removeFilter(filterId);
+  }
+
+  public get column() {
+    return this.dataviewImpl.column;
+  }
+
+  public set column(column: string) {
+    this.dataviewImpl.column = column;
+    this.emit('optionChanged');
```
VictorVelarde: what's the purpouse of 'optionChanged'? who should listen and react to it?

src/lib/dataviews/DataViewWrapperBase.ts
```diff
@@ -0,0 +1,96 @@
...
+    this.dataviewImpl.availableEvents.forEach((e: string) =>
+      this.dataviewImpl.on(e, (args: any[]) => this.emit(e, args))
+    );
+  }
+
+  protected abstract buildImpl(
+    dataSource: Layer | Source,
+    column: string,
+    options: { mode?: DataViewModeAlias },
+    mode: DataViewModeAlias
```
VictorVelarde: why do we have this mode duplicated in options and outside?

src/lib/dataviews/category/Category.ts
```diff
@@ -0,0 +1,57 @@
...
+  protected buildImpl(
+    dataSource: Layer | Source,
+    column: string,
+    options: CategoryOptions,
+    mode: DataViewModeAlias
+  ) {
+    let dataView;
+
+    switch (mode) {
+      case DataViewModeAlias.NON_PRECISE: {
```
VictorVelarde: We still need to decide on 'mode' name, we have to aggree asap on a naming, to apply it internally. After all this time, honestly I would just prefer to use something that matches reality, and internal naming, like 'Local', vs 'Server' mode, and explain via docs its differences on quality, speed and strategy for calculation

src/lib/dataviews/formula/Formula.ts
```diff
@@ -0,0 +1,42 @@
+import { Layer, Source } from '@/viz';
+import { DataViewModeAlias } from '../mode/DataViewModeBase';
+import { AggregationType } from '../../data/operations/aggregation/aggregation';
+import { DataViewLocal } from '../mode/DataViewLocal';
+import { DataViewRemote } from '../mode/DataViewRemote';
+import { DataViewWrapperBase } from '../DataViewWrapperBase';
+import { FormulaImpl } from './FormulaImpl';
+
+export class Formula extends DataViewWrapperBase {
```
VictorVelarde: I would probably remove the 'Base' part from the class

src/lib/dataviews/formula/FormulaImpl.ts
```diff
@@ -0,0 +1,22 @@
...
+import { DataViewImplBase } from '../DataViewImplBase';
+
+export class FormulaImpl<T extends DataViewModeBase> extends DataViewImplBase<T> {
+  public async getData(): Promise<Partial<DataViewData>> {
+    let formulaResponse;
+
+    try {
+      formulaResponse = await this.dataView.formula(this.operation);
+    } catch (error) {
+      this.emit('error', [error]);
```
VictorVelarde, line 12: Should we emit errors or just throw....?

src/lib/dataviews/mode/DataViewLocal.ts
```diff
@@ -0,0 +1,117 @@
+import { Layer } from '@/viz';
+import { AggregationType, aggregate } from '@/data/operations/aggregation/aggregation';
+import { groupValuesByAnotherColumn } from '@/data/operations/grouping';
+import { castToNumberOrUndefined } from '@/core/utils/number';
+import { DataViewModeBase, DataViewData } from './DataViewModeBase';
+import { CartoDataViewError, dataViewErrorTypes } from '../DataViewError';
+
+export class DataViewLocal extends DataViewModeBase {
```
VictorVelarde: same here, I would probably drop the 'Base' part in the parent class

src/lib/dataviews/mode/DataViewRemote.ts
```diff
@@ -0,0 +1,80 @@
...
+import { CartoDataViewError, dataViewErrorTypes } from '../DataViewError';
+
+export class DataViewRemote extends DataViewModeBase {
+  protected dataviewsApi: DataviewsApi;
+
+  constructor(dataSource: Source, column: string, credentials = defaultCredentials) {
+    super(dataSource, column);
+
+    // TODO what about the other sources?
+    const dataset = (dataSource as CARTOSource).value;
```
VictorVelarde, line 26: IMO we should check the type and just throw an 'incompatible source' error

src/lib/index.ts
```diff
@@ -57,11 +57,15 @@ export const viz = {
...
+export const dataview = {
+  Category,
+  Formula
+};
 
-export const data = {
-  ...dataviews
+// carto.widget
+export const widget = {
+  CategoryWidget
```
VictorVelarde: Category: CategoryWidget ?

src/lib/viz/index.ts
```diff
@@ -23,3 +22,6 @@ export {
 
 // Basemap helpers
 export { createGoogleMap, createMap } from './basemap';
+
+// Widgets
+export { Category } from '../widgets';
```
VictorVelarde: IMO it doesn't make sense to export widgets from this index.ts now, as it is a separate namespace (and in the near future we will also extract the 'sources').

src/lib/viz/interactivity/viewport-features/ViewportFeaturesGenerator.ts
```diff
@@ -102,7 +102,7 @@ export class ViewportFeaturesGenerator {
   }
 
   private getSelectedTiles(): ViewportTile[] {
-    if (!this.deckLayer) {
+    if (!this.deckLayer || !this.deckLayer.state) {
```
VictorVelarde, line 105: what's the purpouse of this check?

examples/data/catagory-data-view-all-modes.html
```diff
@@ -0,0 +1,151 @@
+<!DOCTYPE html>
```
VictorVelarde: Let's move dataviews to a new folder in examples, named `dataview` as the namespace, instead of data. Also there is a typo in the file name --> 'catagory'

Don't worry about viewport-features, I have updated it and moved it to another folder in my WIP pull on examples

VictorVelarde:
Nice job Manu. Overall it looks good to me. I made some comments and suggestions, but none of them are blockers. We can talk tomorrow about merging and probably applying my suggestions (and others) in the next PR. For this one I would just change the 'mode' naming. 

Also take note that the 2 'old & simple' dataview examples are now broken (we can take care of them also in the next PR). Maybe you can move them to debug in the meantime.

src/lib/dataviews/category/Category.ts
manmorjim: `mode` renamed:

- `GLOBAL` to `REMOTE`
- `NON_PRECISE` to `LOCAL`
- `VIEWPORT` to `REMOTE_FILTERED` (we have to discuss about this. Not definitive)

src/lib/dataviews/DataViewWrapperBase.ts
manmorjim: I added it to listen when some attribute of the Dataviews has changed in order to fetch the new data and then fire the event `dataUpdated` with the new data retrieved by Maps API. Finally I removed that behaviour and let users of the dataView to fetch the data by themselves so I think I should use `dataUpdated` instead of `optionChanged` and remove this.

By this way the dataViews will fire the event `dataUpdated` when some option has changed in the same way DataView local fires this event when the viewport has changed.

manmorjim: However, with `optionChanged` we were managing through a timeout when to fire `dataUpdate` in order to prevent firing this event multiples times if the options of the dataView has changed in a row. This is, if we set three options of our dataView:

```javascript
categoryDataView.column = column;
categoryDataView.operationColumn = operationColumn;
categoryDataView.operation = operation;
```

With `optionChanged` + `timeout` we are firing the event `dataUpdate` just one time whereas if we remove the `optionChanged` and use `dataUpdate` instead in the example above this event would be firing three times in a row.

src/lib/dataviews/DataViewWrapperBase.ts
manmorjim: nice catch. Removed :+1:

src/lib/dataviews/DataViewWrapperBase.ts
VictorVelarde: If that event is just used internally for debouncing, to trigger less requests, it's ok, but let's try to keep it internal. In fact, I would suggest also extracting the debounce to utils

src/lib/dataviews/formula/FormulaImpl.ts
manmorjim: I copied the [carto.js behaviour](https://carto.com/developers/carto-js/examples/#example-category-widget) but not sure about emitting or throwing

src/lib/viz/interactivity/viewport-features/ViewportFeaturesGenerator.ts
manmorjim: In some examples this method threw me an error because `this.deckLayer` has value but its `state` was null.

If nobody has got that error likely it would be a random error...

src/lib/dataviews/category/category.spec.ts
```diff
@@ -11,7 +11,8 @@ describe('DataView', () => {
         () =>
           new Category(new Layer('fake_source'), 'fake_column', {
             operation: AggregationType.AVG,
-            operationColumn: 'popEst'
+            operationColumn: 'popEst',
+            mode: DataViewModeAlias.LOCAL
```
VictorVelarde: If we're exposing this enum, let's use just DataViewMode. ... without the 'alias'?

VictorVelarde (approved):
LGTM 🚀

PR #17, opened 2020-06-12
Fix CARTOSource instantiation for on-prem

VictorVelarde:
Ensure there is a urlTemplate for the MVTs, even if no subdomains are present. 

The fix is suggested by @Clebal, which detected the issue on the first place, and LGTM.


VictorVelarde:
Just one file affected and 2 commits: 1st commit for the fix, and 2nd for a small refactor

simon-contreras-deel (approved):

PR #18, opened 2020-06-12
Basic examples update

VictorVelarde:
This is a WIP, for a quite detailed review / set of refactors for examples.

Some basic checks: it's relevant for public, can it be modified for a better approach? (or goes to debug), has proper title, description, styles (including responsive)


VictorVelarde:
Indeed examples were already on alpha.2, so just some improvements on basic examples

PR #20, opened 2020-06-15
Change widgets to their own namespace

jesusbotella:
Widgets deserve their own namespace within Web SDK, so here they have it.

From now on, the developers will instantiate: `carto.widgets.Category` instead of `carto.viz.CategoryWidget`


src/lib/index.ts
```diff
@@ -28,6 +25,9 @@ import {
 // Dataviews
 import { CategoryDataView, FormulaDataView, CategorySourceDataView } from './data/dataview';
```
VictorVelarde, line 26: A bit of general discussion on main **classes naming** (as the criteria should apply to this PR and others) and it's quite relevant.

cc/ @manmorjim @oleurud and @borja-munoz 

On one hand the 'type suffix' we are using for some classes (eg. DataView, Widget, even Style), it's not required if we place those classes under a proper namespace... For example `carto.widget.Category` instead of `carto.widget.CategoryWidget`.

But on the other hand, if we apply this principle broadly, Could that be error prone for users, having for example 2 different Category classes (one for the Dataview and one for the Widget)? --> `carto.widget.Category` vs `carto.dataview.Category`?

Wether we go after one option or another, we should be homogeneuous.

Thoughts?

VictorVelarde (changes requested):
Code looks good, but let's discuss all a bit about naming before merging this

src/lib/index.ts
VictorVelarde: If you don't express any preference, @jesusbotella and me will set our option here... 😸

simon-contreras-deel: I also prefer carto.widget.Category over carto.widget.CategoryWidget but having the same class name in the browser is a problem since a user can not import them using an alias, as far as I know, right?

VictorVelarde: well, the user would instantiate a `new carto.widget.Category` or a `new carto.dataview.Category` (assuming we move dataviews into its own namespace, which I think we should). That should cause no issue on the 'technical side'; I was talking more about any "possible confussion" on the naming...

simon-contreras-deel: Yes, you are right, it is not a real problem.  

I think it could be confused in any case (without being a big issue if you know the difference between dataview and widget), I prefer `carto.widget.Category` over `carto.widget.CategoryWidget` just for being shorter.
```
const categoryWidget = carto.widget.Category(...)
const categoryDataview = carto.dataview.Category(...)
```
```
const categoryWidget = carto.widget.CategoryWidget(...)
const categoryDataview = carto.dataview.CategoryDataview(...)
```

VictorVelarde: As this discussion is bigger, IMO it deserves more time and a better place, so @jesusbotella I'm unlockling this for merging with the current approach (option 1: `carto.widget.Category`). Linking here our CH story: https://app.clubhouse.io/cartoteam/story/84574/product-refactor-api

VictorVelarde (approved):
🚀  Let's go for it!

PR #21, opened 2020-06-15
small styles fixes from review in cartoframes

simon-contreras-deel:
https://app.clubhouse.io/cartoteam/story/80830/styles-review-in-cartoframes-by-chemita

- [x] (1) review default styles (without helpers)
    It can not be solved completely since we are not supporting functions from cartoCSS. I have decreases a bit points size and stroke color opacity.

- [x] (2) color continuous: the distance between points is different
    The distance was right. I have decreased points size. Again, in CF we have a dynamic ramp we are not supporting here, so I have used a smaller but constant value

- [x] (3) size category with points: in VL color is orange! Also review opacity
    Done 

- [x] (4) size continuous with points: review stroke
    Same as before with ramp function.

- [x] (5) color continuous with polygons: stroke
    Added stroke just to polygons

### DIFF
--- a/src/lib/viz/style/default-styles.ts
+++ b/src/lib/viz/style/default-styles.ts
@@ -2,12 +2,12 @@ const NULL_COLOR = '#ccc';
 const OTHERS_COLOR = '#777';
 const NULL_SIZE = 0;
 const STROKE_WIDTH = 0.5;
-const STROKE_COLOR = '#22222288';
+const STROKE_COLOR = '#22222244';
 
 export const defaultStyles: any = {
   Point: {
     color: '#EE4D5A',
-    size: 10,
+    size: 8,
     opacity: 1,
     strokeColor: STROKE_COLOR,
     strokeWidth: STROKE_WIDTH,
--- a/src/lib/viz/style/default-styles.ts
+++ b/src/lib/viz/style/default-styles.ts
@@ -9,7 +9,7 @@ export const defaultStyles: any = {
     color: '#EE4D5A',
     size: 8,
     opacity: 1,
-    strokeColor: STROKE_COLOR,
+    strokeColor: '#FFF',
     strokeWidth: STROKE_WIDTH,
     nullColor: NULL_COLOR,
     othersColor: OTHERS_COLOR,
--- a/src/lib/viz/style/default-styles.ts
+++ b/src/lib/viz/style/default-styles.ts
@@ -2,7 +2,7 @@ const NULL_COLOR = '#ccc';
 const OTHERS_COLOR = '#777';
 const NULL_SIZE = 0;
 const STROKE_WIDTH = 0.5;
-const STROKE_COLOR = '#22222244';
+const STROKE_COLOR = '#FFFFFF';
 
 export const defaultStyles: any = {
   Point: {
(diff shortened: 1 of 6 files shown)


VictorVelarde:
Overall it looks good. But before merging, one comment about our default **strokeColor**.

I haven't checked it broadly, but considering a point layer, I think we should change it (not sure if just for the points, or also for polygons). Using a white, instead of the current 'gray', give us a better match with CARTO VL (I would say a perfect match)

This is the result with the change:

image

I have added a commit, for you to have a look. Feel free to adapt or revert if you have a clear reason

src/lib/viz/style/default-styles.ts
```diff
@@ -2,14 +2,14 @@ const NULL_COLOR = '#ccc';
 const OTHERS_COLOR = '#777';
 const NULL_SIZE = 0;
 const STROKE_WIDTH = 0.5;
-const STROKE_COLOR = '#22222288';
+const STROKE_COLOR = '#22222244';
```
VictorVelarde: This is the color I have doubts about. See my other comment

src/lib/viz/style/default-styles.ts
```diff
@@ -2,14 +2,14 @@ const NULL_COLOR = '#ccc';
...
 
 export const defaultStyles: any = {
   Point: {
     color: '#EE4D5A',
-    size: 10,
+    size: 8,
     opacity: 1,
-    strokeColor: STROKE_COLOR,
+    strokeColor: '#FFF',
```
VictorVelarde, line 12: My proposal, feel free to push it up as the new STROKE_COLOR if it fits

simon-contreras-deel: I think the objective is to fit with CARTOframes style helpers, not with VL ones, right?

VictorVelarde (approved):
LGTM. Just consider my stroke_color comments before merging

src/lib/viz/style/default-styles.ts
VictorVelarde: And overall, especially on style helpers, they match and they are great. But in this particular  case of just defaults, I think that a grey border is clearly worse than a white one for points, as it doesn't allow to see clearly the circles. Honestly I'm a bit surprised that they use that in CF. We can discuss a bit with them if we need to

VictorVelarde: Helpers were born in CF, so it's natural to me to take them as a reference, but IMO defaults should work in a broader context

VictorVelarde: In any case, don't take me too seriously on this. If they are pretty sure about that default, we can just respect it and move on. In that case, just revert my commit and let's move on

simon-contreras-deel: As we have talked before, we are going to use a white stroke. I am going to add a commit with that.

josema-xyz:
The modifications are great overall, but... it looks like the white isn't the best solution for every case (IMO). But I shouldn't decide about that, so I'm going to approve the PR.

image

image

josema-xyz (approved):
:raised_hands:

PR #22, opened 2020-06-15
ensuring proper related style properties

simon-contreras-deel:
https://app.clubhouse.io/cartoteam/story/84011/review-ensureproperpropstyles

We are ensuring related style properties twice. This PR remove one of them. I think we need to do it just before sending it to the source or it could be overwritten


VictorVelarde (approved):
Yes, I think `ensureRelatedStyleProps` is equivalent

manmorjim (approved):

manmorjim:
:ok_hand:

simon-contreras-deel:
Branch rebased with develop. Vercel is failing but since we know it is not a real problem, I am going to merge it

PR #23, opened 2020-06-15
Relaunch map instantiation if new metadata required from styles

VictorVelarde:
If a new style is set, with Layer.setStyle, that requires a new column, we're not reinstantiating the map (vector map in maps api, aka source.init), but we should, or it will crash. Something similar can happen if we require more columns in a popup after the first instantiation


simon-contreras-deel:
**style is changed after layer was added to deck**:
- [x] create new deck layer

**popup is added after layer was added to deck**:
- [x] layer calls source.init after setting a popup
- [x] Sources does nothing if the fields do not change or send a warning (instead of the current error in CARTOSource)

simon-contreras-deel:
@VictorVelarde it is ready to be reviewed

simon-contreras-deel:
Again in review. An example file has been created to help in validation `examples/_debug/source/reinstantiation.html`. The idea is to play adding styles, popups and so after the layer is added to deck.

VictorVelarde:
I'll see the conflict and CR

VictorVelarde (commented):
Nice work and testing.

Please @simon check the conflict and also update the branch examples to beta.1. Then IMO we're ready to go (but you'll have to approve it, as I was the original creator... 😄 )

simon-contreras-deel:
Rebasing it...

simon-contreras-deel (approved):
Self-approve :)

PR #25, opened 2020-06-16
DataView modes with filters

manmorjim:


src/lib/dataviews/category/Category.ts
```diff
@@ -23,7 +25,8 @@ export class Category extends DataViewWrapper {
       }
 
       default: {
-        dataView = new DataViewRemote(dataSource as Source, column);
+        dataView = new DataViewRemote(dataSource as Layer, column);
+        dataView.addFilter(`VIEWPORT_FILTER_${uuidv4()}`, BuiltInFilters.VIEWPORT);
```
jesusbotella, line 29: I guess that the switch is the way we have to differentiate between all data view types, right?

So we would just apply this type of filter to the remote data view when no type is provided? I'm not sure to fully grasp this switch split.

src/lib/maps/Client.ts
```diff
@@ -78,7 +78,11 @@ export class Client {
    * @param layergroup
    * @param options
    */
-  public async dataview(layergroup: any, dataview: string, categories?: number) {
+  public async dataview(
+    layergroup: any,
```
jesusbotella: I think we should have types for this anywhere, because we are using them in MapConfig

src/lib/maps/Client.ts
```diff
@@ -89,9 +93,21 @@ export class Client {
...
+        }
+
+        if (optionToEncode) {
+          const encodedOption = encodeParameter(option, optionToEncode);
+          parameters.push(encodedOption);
+        }
+      });
     }
 
     const getUrl = `${url.https}?${parameters.join('&')}`;
```
jesusbotella: It's not part of this specific PR but we could use a safer method to generate the URL with URLSearchParams or something like that.

src/lib/maps/Client.ts
```diff
@@ -263,3 +279,53 @@ export interface MapInstance {
...
+   */
+  aggregationColumn?: string;
+
+  /**
+   * Bounding box to filter data
+   */
+  bbox?: number[];
+}
+
+export enum AggregationType {
```
jesusbotella: I think we have them already declared in another modules, and we should avoid having enumerations with the same values.

jesusbotella (approved):
I think it looks good. I gave it an overall look, and I didn't think of reusing the filters like this but turns out to be a good approach and reutilization.

src/lib/dataviews/category/Category.ts
manmorjim: You're right. ATM the default mode is REMOTE so what about putting REMOTE as default? This is:

```typescript
    switch (mode) {
      case DataViewCalculation.LOCAL: {
        dataView = new DataViewLocal(dataSource as Layer, column);
        break;
      }

      case DataViewCalculation.REMOTE_FILTERED: {
        dataView = new DataViewRemote(dataSource as Layer, column);
        dataView.addFilter(`VIEWPORT_FILTER_${uuidv4()}`, BuiltInFilters.VIEWPORT);
        break;
      }

      default: {
        dataView = new DataViewRemote(dataSource as Source, column);
        break;
      }
    }
```

src/lib/maps/Client.ts
manmorjim: I didn't know about URLSearchParams! more clear and simpler! :+1:

src/lib/maps/Client.ts
manmorjim: nice catch!

PR #26, opened 2020-06-19
Formula widget

jesusbotella:
This PR adds Formula Widget to the current set of widgets within Web SDK.

To initialize the widget within the DOM with DataView data, you just need to:
```js
const formulaDataview = new carto.data.FormulaDataView(countriesLayer, 'cartodb_id', {
  operation: 'count',
  operationColumn: 'cartodb_id'
});

new carto.widgets.Formula('as-formula-widget', formulaDataview)
```


examples/widgets/formula.html
```diff
@@ -0,0 +1,75 @@
...
+    async function initialize() {
+      carto.setDefaultCredentials({ username: 'public' });
+
+      const deckMap = carto.viz.createMap();
+
+      const countriesLayer = new carto.viz.Layer('ne_50m_admin_0_countries');
+      await countriesLayer.addTo(deckMap);
+
+      // Formula Widget
+      const formulaDataview = new carto.data.FormulaDataView(countriesLayer, 'cartodb_id', {
```
VictorVelarde: I think we'll have to change it to `carto.dataview.Formula`, after @manmorjim current PR. Also, as we will probably change the 'count', to avoid the need of a param, maybe we can just use 'sum' or another formula.

Regarding to the airship panel, there is right now an example you did for Category Widget that I changed a bit, to add the title above the widget. You can probably copy that layout for this, and we'll get a more homogeneous look, and a basic responsive panel

jesusbotella: I have just pushed the changes.

Regarding what we talked yesterday of renaming `Category` class (or any other that has the same issue) to `CategoryWidget`, I think we should do it as soon as we can. I have had to add another alias for FormulaDataview in the Formula (FormulaWidget) class. We are gonna end up with alias everywhere.

VictorVelarde (approved):
LGTM. Feel free to merge, once we have merged and published https://github.com/CartoDB/airship/pull/696

PR #27, opened 2020-06-22
Add custom source to allow users to pass a custom Maps API instantiat…

sclebal:
…ion as carto.viz.Layer argument.

I'm pretty sure that it isn't good but I want to open this PR to "start" the process to add this feature in Web SDK.


VictorVelarde:
I just had a quick look on this, but a first question: how do you instantiate your 'custom' source? could you post a little snippet? I understand you will eventually have a CARTOSource, but instantiated externally... So basically your 'custom' is a CARTOSource where you want to omit the instantiation. right?

sclebal:
@VictorVelarde I've already uploaded a first basic example to show you how it works in `/_debug/source/custom.html` file. I've used a front request just for example purpose, but the idea is to have that request in the back.

The idea is the following:

In a certain project you have an API where the SQL queries (the ones to get the data to visualize) are stored. The front doesn't know anything about SQL, it just calls to a certain endpoint `https://example-api.com/map/shops/` with some params and the API itself call the Maps API with the config necessary to get all the correct metadata. You cannot have CARTOSource in a Python backend, that's why I avoid that.

What I wanted to achieve is to have the same CARTOSource as always but "extracting" the logic of the Maps API request to allow users to make their own requests and take care of them. The others parts should be the same.

> So basically your 'custom' is a CARTOSource where you want to omit the instantiation. right?

Yes, I think that's the easiest way.

simon-contreras-deel:
Adding the do not merge yet label. I have created an issue to discuss about the source problem

VictorVelarde:
Thx Sergio for your work on this, but after the research made it looks like 'hiding the instantiation' is not really required. So I'm closing this PR.

PR #29, opened 2020-06-22
Update create-map.ts

juandjara:


src/lib/viz/basemap/create-map.ts
```diff
@@ -55,9 +55,12 @@ export function createMap(options: DeckGLMapOptions = DEFAULT_OPTIONS) {
 
   const deckMap = new (window.deck.DeckGL as any)({
     mapStyle: CartoMapStyle[chosenOptions.basemap.toUpperCase() as keyof typeof CartoMapStyle],
-    initialViewState: chosenOptions.view,
+    viewState: chosenOptions.view,
     container: chosenOptions.container,
-    controller: true
+    controller: true,
+    onViewStateChange: ({ viewState }) => {
+      deckMap.setProps({ viewState })
```
jesusbotella: Not sure about this. We'd be supplying the very same viewState to the map whenever it changes. Seems like an unnecessary overhead.

VictorVelarde:
From the docs: https://deck.gl/#/documentation/developer-guide/views-and-projections?section=using-a-view-class

If **initialViewState** is provided, deck.gl automatically tracks the view states of interactive views (used as a "stateful" component):

If you need to manage and manipulate the view state outside of deck.gl, you may do so by providing an external **viewState** prop (used as a "stateless" component). In this case, you also need to listen to the onViewStateChange callback and update the viewState object yourself:

juandjara:
@VictorVelarde thanks for the link. That describes exactly what we want to do, "manage and manipulate the view state outside of deck.gl"

src/lib/viz/basemap/create-map.ts
simon-contreras-deel: I agree

simon-contreras-deel: I think we would be forcing it to change state when "the state has changed". I mean, creating a never-ending state update loop, don't you think?

juandjara: As it says in the comment from Victor below, the map does not update itself with pan and zoom events if `initialViewState` is not set. What it does is just send a `onViewStateChange` event with the new viewState so you can update it with `setProps`

sclebal:
@VictorVelarde @jesusbotella @oleurud Without this change when I change the view the map remain static. This PR is necessary to avoid it. Thanks @juandjara.

VictorVelarde:
I'm working on this

VictorVelarde (approved):
I have made just small adjustments, and I added a debug example to test fit-bounds

PR #30, opened 2020-06-22
Fix DataView instantiation in remote mode with a layer

manmorjim:
Checks if the parameter is a layer to get its source


VictorVelarde (approved):
LGTM

PR #31, opened 2020-06-22
[FIX] DataView with an SQL query

manmorjim:
This fixes:
- Using the name "dataview" if the source is an SQL instead of the full SQL
- Using the SQL query as is instead of wrapping the query with a `SELECT * FROM` statement

### DIFF
--- a/src/lib/maps/Client.ts
+++ b/src/lib/maps/Client.ts
@@ -65,7 +65,7 @@ export class Client {
           type: 'source',
           id: `${source}_${Date.now()}`,
           params: {
-            query: `SELECT * FROM ${source}`
+            query: type === 'sql' ? source : `SELECT * FROM ${source}`
           }
         }
       ],
--- a/src/lib/maps/Client.ts
+++ b/src/lib/maps/Client.ts
@@ -1,3 +1,4 @@
+import { uuidv4 } from '@/core/utils/uuid';
 import { Credentials } from '../core/Credentials';
 import errorHandlers from './errors';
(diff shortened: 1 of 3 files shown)


src/lib/maps/MapsDataviews.ts
```diff
@@ -60,6 +60,18 @@ export class MapsDataviews {
...
   }
+
+  private getDataviewName(): string {
+    let dataviewName = this._source;
+
+    if (dataviewName.search(' ') > -1) {
+      dataviewName = 'dataview';
+    }
+
+    dataviewName = `${dataviewName}_${Date.now()}`;
```
jesusbotella: I thunk we could use UUID id for that, as we used it for the widgets as well.

src/lib/maps/Client.ts
```diff
@@ -65,7 +65,7 @@ export class Client {
           type: 'source',
           id: `${source}_${Date.now()}`,
           params: {
-            query: `SELECT * FROM ${source}`
+            query: type === 'sql' ? source : `SELECT * FROM ${source}`
```
jesusbotella, line 70: DIdn't we have this already?

manmorjim: Well, we are using something similar at [Client.ts](https://github.com/CartoDB/web-sdk/blob/ad24d4ef51c5d4e4392d1c59e61800c486676ff3/src/lib/maps/Client.ts#L44) but with a MapOptions parameter

jesusbotella (approved):

PR #34, opened 2020-06-24
[WIDGET] Category Widget (II) - with Filter

manmorjim:
Allow pass a `Layer` as parameter in `REMOTE` mode. By this way filters can be applied using the `addFilter` method of
Layer.

If a Source is provided to `REMOTE_FILTERED` mode an error is thrown.


jesusbotella (approved):
It looks good.

About the thing you say in the PR description, how do you pass filters to the source from the layer? Do you add them to the source when they are added to the layer?

manmorjim:
> how do you pass filters to the source from the layer? Do you add them to the source when they are added to the layer?

AFAIK the widget adds the filter `in: ['category']` to the DataView which adds it to the dataSource through the method `addFilter`. In `REMOTE` mode the user can create the DataView from a `Layer` or a `Source`.

If the user passes a `Layer` as parameter the DataView will do `layer.addFilter('id', {'column': {in: [''category']}})` so if the parameter was a `Source` then the error `Method not implemented` will be thrown because this method is not implemented in this class.

Not sure if this answers your question :sweat_smile:

PR #35, opened 2020-06-24
[Fix] debounce attribute assignments

manmorjim:
The attribute assignment statements are inside the debounce function which are being cancelled by others set methods. By this way if the user runs the following piece of code:

```typescript
dataview.column = column;
dataview.operation = operation;
dataview.limit = limit;
```

Only the option `limit` would be set.

This fixes that, by putting the attribute assignments outside the debounce method.


VictorVelarde (approved):
LGTM

PR #36, opened 2020-06-24
[Dataview] Aggregated count

manmorjim:
CH: https://app.clubhouse.io/cartoteam/story/85912/dataview-aggregated-count

### DIFF
--- a/src/lib/dataviews/mode/DataViewLocal.ts
+++ b/src/lib/dataviews/mode/DataViewLocal.ts
@@ -1,9 +1,10 @@
 import { Layer } from '@/viz';
 import { AggregationType, aggregate } from '@/data/operations/aggregation/aggregation';
-import { groupValuesByAnotherColumn, groupValues } from '@/data/operations/grouping';
+import { groupValuesByColumn } from '@/data/operations/grouping';
 import { castToNumberOrUndefined } from '@/core/utils/number';
 import { DataViewMode, DataViewData } from './DataViewMode';
 import { CartoDataViewError, dataViewErrorTypes } from '../DataViewError';
+import { CategoryElement } from '../category/CategoryImpl';
 
 export class DataViewLocal extends DataViewMode {
   constructor(dataSource: Layer, column: string) {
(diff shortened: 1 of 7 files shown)


src/lib/dataviews/mode/DataViewLocal.ts
```diff
@@ -53,15 +53,19 @@ export class DataViewLocal extends DataViewMode {
...
-      this.column
-    );
+    let groupedValues;
+
+    if (operationColumn) {
+      const sourceData = await this.getSourceData([operationColumn]);
+      groupedValues = groupValuesByAnotherColumn(sourceData, operationColumn, this.column);
+    } else {
+      const sourceData = await this.getSourceData();
+      groupedValues = groupValues(sourceData, this.column);
```
jesusbotella: Not sure about this approach. It's true that the grouping is kinda different because this one returns a number, but I would simplify that until we see a performance or memory problem.

It would use the same function to group them both, whether operation is count or not.

So, I would perform these changes:
- Remove `groupValues` function.
- Renaming `groupValuesByAnotherColumn` to `groupValuesByColumn`
- Refactor this code so that could be used in both cases like:

```js
const sourceData = await this.getSourceData([ operationColumn || this.column ]);
const { groups, nullCount } = groupValuesByColumn(
  sourceData,
  operationColumn || this.column,
  this.column
);
```

src/lib/dataviews/mode/DataViewLocal.ts
```diff
@@ -82,18 +86,26 @@ export class DataViewLocal extends DataViewMode {
...
-  const numberFilter = function numberFilter(value: number | undefined) {
-    return Number.isFinite(value as number);
-  };
+function createCategory(name: string, data: number[] | number, operation: AggregationType) {
+  let categoryValue;
 
-  const filteredValues = data
-    .map(number => castToNumberOrUndefined(number))
-    .filter(numberFilter) as number[];
+  if (operation === AggregationType.COUNT) {
```
jesusbotella: I think we should call aggregate at all times in this code to make it standard given that we have the method already implemented.

So I would leave the method as is, but avoiding to filter values if operation is count.

```js
let categoryValues = data;
const shouldFilterValues = operation !== AggregationType.COUNT;

if (shouldFilterValues) {
  const numberFilter = function numberFilter(value: number | undefined) {
    return Number.isFinite(value as number);
  };

  categoryValues = (data as number[])
    .map(number => castToNumberOrUndefined(number))
    .filter(numberFilter) as number[];
}

return {
  name,
  value: aggregate(filteredValues, operation)
}
```

jesusbotella (changes requested):
I have pointed out some things that we can discuss. I think we should make the approach more general and not so focused on the count operation.

jesusbotella (approved):
It looks good!

Why do we sort categories now? Displaying them in alphabetical order should be a presentation issue, not a dataview issue.

manmorjim:
> Why do we sort categories now? Displaying them in alphabetical order should be a presentation issue, not a dataview issue.

Well, I've just realized the remote widgets are displaying the categories sorted by alphabetical order meanwhile the local widget was displaying the categories in a random order which looks weird.

But you're right, that is a presentation issue. Do you think I should revert the sort commit and address that issue in a different ticket? that issue should be fixed at widget level?

jesusbotella:
Let's leave it like that for now. Given that we have other things to do, this is not super important.

PR #37, opened 2020-06-24
Add position property to Popup constructor

sclebal:


manmorjim (commented):
LGTM :ok_hand: 

Great improvement! This allows adapting popup to both left and right map edges. However, I miss `top` and `bottom` positions in order to allow popups to be adapted to whole viewport edges.

By this way, if the popup is close to a viewport corner this would be adapted. E.g. if the popup is close to the upper right corner then the position would be `left-bottom`:

image

I also miss a method to set the position for a popup already instantiated.

sclebal:
@manmorjim check what I've just uploaded.

manmorjim:
Nicee! :100: 

LGTM :ok_hand:

manmorjim (approved):

VictorVelarde:
I see the result (and the potential for a future-better auto-positioning), but @Clebal could you please describe a bit your use case now, with this addition?

sclebal:
@VictorVelarde There isn't an use case, it's something about design. Designers what the popup in the top-center and I need a way to achieve it.

VictorVelarde:
Ok, that's enough for me. It's a fair design request.

VictorVelarde:
@manmorjim please, have a last look and let's merge this one!

PR #38, opened 2020-06-24
Sergioclebal/add new styles and style function

sclebal:


VictorVelarde:
Could you please give us a simple example?

sclebal:
@VictorVelarde I've added a basic example.

VictorVelarde:
We want to think about this part with more calm, so we won't be able to tackle it in the current iteration

PR #32, opened 2020-06-22
getViewportFeatures and GeoJSON

simon-contreras-deel:
https://app.clubhouse.io/cartoteam/story/86624/dataview-getviewportfeatures-working-in-geojson

Validation examples:
- https://web-sdk-git-oleurud-ch86624dataview-getviewportf-c2f29f.carto-frontend.vercel.app/examples/_debug/source/polygons-geojson-viewport-features
- https://web-sdk-git-oleurud-ch86624dataview-getviewportf-c2f29f.carto-frontend.vercel.app/examples/_debug/source/points-geojson-viewport-features

PR #40, opened 2020-06-25
dataviews and geojson

simon-contreras-deel:
https://app.clubhouse.io/cartoteam/story/86981/dataviews-calculations-on-geojson-source

- [x] local (already working)
- [x] remote filtered: using the local one
- [x] remote: using the local one without viewport

Needs https://github.com/CartoDB/web-sdk/pull/32 first

### DIFF
--- a/examples/_debug/widgets/geojson-category-data-view-all-modes.html
+++ b/examples/_debug/widgets/geojson-category-data-view-all-modes.html
@@ -43,9 +43,9 @@
             </p>
             <p>
               <span class="as-caption">Operation</span>
-              <as-dropdown default-text="Operation" can-clear="false" selected-option="count"></as-dropdown>
+              <as-dropdown default-text="Operation" can-clear="false" selected-option="sum"></as-dropdown>
             </p>
-            <p>
+            <p id="operationColumnSection">
               <span class="as-caption">Operation column</span>
               <input id="operationColumn" class="as-input" type="text" value="pop_est" />
             </p>
@@ -74,6 +74,7 @@
   <script src="/dist/umd/index.min.js"></script>
 
   <script>
+    const operationColumnSection = document.getElementById('operationColumnSection');
     const operationElem = document.querySelector('as-dropdown');
     operationElem.options = [
       { text: 'SUM', value: 'sum' },
@@ -82,6 +83,9 @@
       { text: 'MAX', value: 'max' },
       { text: 'MIN', value: 'min' }
     ];
+    operationElem.addEventListener('optionChanged', function (event) {
+      operationColumnSection.style.display = (event.detail === 'count') ? 'none' : '';
+    });
     const widgetRemoteElement = document.getElementById(
       'category-widget-remote'
       );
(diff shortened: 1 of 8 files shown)


manmorjim (commented):
LGTM! :heavy_check_mark: 

However I'd like hearing your opinion about this (cc  @jesusbotella ):

What do you think about using the same method `getFeatures` for MVT and GeoJSON? This would be, renaming the method  `getViewportFeatures` from `Layer` to `getFeatures` and both `CARTOSource` and `GeoJSONSource` implementing this method with the same signature.

As I see it, the method `getViewportFeatures` within a `Source` is equivalent to adding a spatial filter with the current viewport bounds and calling the method `getFeatures`.
The only difference between `CARTOSource` and `GeoJSONSource` is that the method `getFeatures` of `CARTOSource` will return the same features even though a viewport filter had been applied.

With this, the flag `useViewport` is not needed anymore and the only difference between GeoJSON Layers and MVT Layers would be GeoJSON Layers always use a `DataViewLocal` instance.

simon-contreras-deel:
Yes, I was expecting from your side some ideas like the one you proposed.

examples/_debug/widgets/geojson-category-data-view-all-modes.html
```diff
@@ -0,0 +1,175 @@
...
+
+      const geojson = await fetch('https://public.carto.com/api/v2/sql?api_key=default_public&q=select* from ne_50m_admin_0_countries&format=geojson').then(response => response.json())
+      const countriesLayer = new carto.viz.Layer(geojson);
+      await countriesLayer.addTo(deckMap);
+
+      categoryRemote = new carto.dataview.Category(
+        countriesLayer.source,
+        'continent',
+        { operation: 'count', operationColumn: 'pop_est', mode: 'remote' }
+      );
```
VictorVelarde: Some comments, also for @manmorjim. Can't we already pass just a `countriesLayer` instead of countriesLayer.source` for this?

Also, the 'count' examples are maybe not the best ones, as it doesn't really need an `operationColumn' (that would probably confuse our "future us")

VictorVelarde:
I think it's important to keep the `getViewportFeatures` in the Layer, as it is very frequent and useful. But I guess that's not incompatible with using getFeatures (without the 'viewport') in the dataviews, if we can get that with filters. So +1 for that part of the refactor

VictorVelarde:
I have read again that part of the code, and it looks like that could be incompatible, without renaming Layer.getViewportFeatures...

VictorVelarde (approved):
LGTM. I'm ok with the current approach, keeping viewportFeatures in Layer (even at the cost of adding a new (default) parameter to DataView local... If you find another clear approach, ping me for a new revision, although I would prefer to do that in a new PR

examples/_debug/widgets/geojson-category-data-view-all-modes.html
manmorjim: > Can't we already pass just a `countriesLayer` instead of countriesLayer.source` for this?
 
Yes, we can :smiley:

VictorVelarde (changes requested):
Ey, I have just seen we're missing Formula... don't we need any adaptation to it too?

examples/_debug/widgets/geojson-category-data-view-all-modes.html
simon-contreras-deel: I have updated this example from the original one which is using SUM and layers

simon-contreras-deel:
>  we're missing Formula

Good catch, I forgot it! I have added it, ready to review again

VictorVelarde (approved):
LGTM

VictorVelarde:
I'm gonna merge this, to unlock other tasks

PR #43, opened 2020-06-29
[Refactor API] r(b): carto.source

manmorjim:
- Creating the namespace `carto.source`
- Dropping 'Source suffix' just in public API:
  - `carto.source.DO`
  - `carto.source.GeoJSON`
  - `carto.source.SQL`
  - `carto.source.Dataset`


src/lib/index.ts
```diff
@@ -43,8 +43,15 @@ export const auth = {
   setDefaultCredentials
 };
 
+// carto.source
+export const source = {
+  DO: DOSource,
```
VictorVelarde: We won't expose this DO yet, you can keep it commented and with a note

src/lib/source/GeoJSONSource.ts
```diff
@@ -1,4 +1,4 @@
-import { GeoJSON, GeoJsonGeometryTypes, Feature } from 'geojson';
+import { GeoJSON, Feature, GeoJsonGeometryTypes } from 'geojson';
```
VictorVelarde: It's curious how they use GeoJSON and GeoJson in the same lib... 😄

VictorVelarde (commented):
Overall, it looks really good to me. I feel like the result is more clear and the 3 commits approach made the review easier, to thx for it. Let's wait a bit for geojson and histogram dataviews to be merged, and then I'll approve this

src/lib/source/GeoJSONSource.ts
manmorjim: yup, I thought the same... _why did they do that?_ :shrug:

VictorVelarde (approved):
👌 . Nice, let's go ahead!

PR #44, opened 2020-06-29
Histogram DataView

jesusbotella:
This PR implements Histogram DataView to retrieve data aggregated by bins.


VictorVelarde:
This requires https://github.com/CartoDB/web-sdk/pull/45., so they both need to be merged together

src/lib/dataviews/histogram/Histogram.ts
```diff
@@ -0,0 +1,39 @@
...
+        dataView = new DataViewRemote(dataSource as Source, column);
+        break;
+      }
+
+      default: {
+        dataView = new DataViewRemote(dataSource as Layer, column);
+        dataView.addFilter(`VIEWPORT_FILTER_${uuidv4()}`, BuiltInFilters.VIEWPORT);
+        break;
+      }
+    }
```
VictorVelarde: @jesusbotella I think we will need here a similar approach as this one (still WIP) to make it work with GeoJSON, right? https://github.com/CartoDB/web-sdk/blob/193f57b518b6dd20af1e755369090a7fce24ca78/src/lib/dataviews/category/Category.ts

VictorVelarde (changes requested):
Overall, it looks good to me. But I think we're missing GeoJSON sources, so we'll need to add a bit more of work on that. It would also be good to have some specs, eg for the client-side code (the histogram calculations from DataViewLocal)

src/lib/dataviews/histogram/Histogram.ts
jesusbotella: Yes, kinda, we should simplify the code in there, but yes, we would need something like that.

VictorVelarde: We will have that repetition also in Formula (@oleurud is going to add that to his PR). A candidate for further refactor in the future, but step by step

src/lib/viz/utils/check.ts
```diff
@@ -0,0 +1,9 @@
+import { Layer, Source } from '@/viz';
+import { GeoJsonSource } from '../sources';
+
+export function isGeoJSONSource(dataSource: Layer | Source) {
```
simon-contreras-deel: This function already exists in  `src/lib/dataviews/utils.ts`

src/lib/dataviews/histogram/Histogram.ts
```diff
@@ -13,27 +14,34 @@ export class HistogramDataView extends DataViewWrapper {
...
+      const useViewport = false;
+      return new DataViewLocal(dataSource as Layer, column, useViewport);
     }
 
-    this.dataviewImpl = new HistogramImpl(dataView, options);
+    return new DataViewRemote(dataSource as Source, column);
+  },
+
+  [DataViewCalculation.REMOTE_FILTERED](dataSource: Layer | Source, column: string) {
+    const dataView = new DataViewRemote(dataSource as Layer, column);
```
simon-contreras-deel, line 47: Don't you need to avoid adding the filter in geoJSON case?

jesusbotella: Yes, you are right, I missed that second if in the original code. We need to avoid adding the filter and creating a remote DataView which makes no sense.

I was wondering it it would be better to throw an error when using a GeoJSON source with any unsupported mode, like `REMOTE_FILTERED` mode rather than "fixing it" automagically.

src/lib/viz/utils/check.ts
jesusbotella: Oops! I missed that one.

VictorVelarde:
@oleurud please approve it if you consider it is ok from the GeoJSON perspective

simon-contreras-deel (approved):
I think it should work fine with geoJSON

VictorVelarde (approved):
Go ahead @jesusbotella !

jesusbotella:
~Our goal should be to properly position the popup within the screen in those edge cases. So even if the developers need to add the `overflow:hidden` rule to CSS (like Mapbox users do) it should not limit us when taking the decision, because they should add it regardless positioning is fixed or not.~

VictorVelarde:
@jesusbotella  I think your prev. comment goes to https://github.com/CartoDB/web-sdk/pull/51

jesusbotella:
Oh! :(
You're right, I think I posted it to the wrong tab :(

VictorVelarde:
Could we merge this now or is anything pending?

jesusbotella:
Yes, I thought I had merged it already, but as the branch was out-of-date I updated the branch and waited until tests passed.

As there are more commits in develop I'm gonna merge it before updating the branch again.

PR #45, opened 2020-06-29
Histogram Widget

jesusbotella:
This PR adds Histogram Widget with `WITHIN` filter to hide filtered features in the map


examples/widgets/histogram.html
```diff
@@ -0,0 +1,98 @@
...
+  <script src="https://api.tiles.mapbox.com/mapbox-gl-js/v1.10.0/mapbox-gl.js"></script>
+  <script src="https://unpkg.com/deck.gl@8.2.0-beta.1/dist.min.js"></script>
+  <script src="https://libs.cartocdn.com/airship-components/v2.3/airship.js"></script>
+  <script src="/dist/umd/index.min.js"></script>
+
+  <script>
+    const widgetElement = document.querySelector('as-category-widget');
+
+    async function initialize() {
+      carto.setDefaultCredentials({ username: 'public' });
```
VictorVelarde: This should be adapted, now we have the first refactor merged on carto.auth

examples/widgets/histogram.html
```diff
@@ -0,0 +1,98 @@
...
+        bins: 10, start: 1000000, end: 127078679, mode: 'REMOTE_FILTERED'
+      });
+      const localHistogramWidget = new carto.widget.Histogram(document.querySelector('.local'), localDataview);
+
+
+      const remoteDataview = new carto.dataview.Histogram(countriesLayer, 'pop_est', {
+        bins: 10, start: 1000000, end: 127078679, mode: 'LOCAL'
+      });
+      const remoteHistogramWidget = new carto.widget.Histogram(document.querySelector('.remote'), remoteDataview);
+    }
```
VictorVelarde: I find this naming a bit confusing. Shouldn't the first `localDataview` be called something in the line of `viewportHistogramDataview` (& viewportHistogramWidget)?. 

Also. using 'LOCAL' for the `remoteDataview` feels bizarre

src/lib/widgets/histogram/Histogram.ts
```diff
@@ -0,0 +1,55 @@
...
+      this.dataView.addFilter(this.widgetUUID, { within: selectionData.selection });
+    });
+
+    await this.updateData();
+  }
+
+  protected async updateData() {
+    const data = await this.dataView.getData();
+    const histogramWidget = this.element as HTMLAsHistogramWidgetElement;
+    histogramWidget.data = (data as HistogramDataViewResult).bins;
```
VictorVelarde: `HistogramDataViewData` is more homogeneous?

VictorVelarde (changes requested):
OK. Some comments about 'naming' on examples, but LGTM. I won't approve it yet (we could), just to wait a bit for the geojson-related PRs to go first

examples/widgets/histogram.html
jesusbotella: Yes, I didn't review this before committing. I'm gonna change this.

PR #50, opened 2020-06-30
[Popups] Popup adjust size & x button

manmorjim:
- Close button styles inline
- Removing css styles from examples


VictorVelarde (approved):
LGTM

PR #51, opened 2020-06-30
[Popups] Scroll-bar on map when popup 'pushes-it' 

manmorjim:
Prevent scroll-bar appearing by setting the overflow style to `hidden` when the map is created


jesusbotella:
I am not sure whether we need to attach this behavior to the Popup or not. I think `overflow` property is meant to be on the map container, rather than searching Popup's container and adding the `overflow: hidden` to it.

We could probably check out how Mapbox or any other map libraries with popups do this.

manmorjim:
> We could probably check out how Mapbox or any other map libraries with popups do this.

I took a look at mapbox and its map container has the CSS rule `overflow: hidden` so we have several options:

- Adding this rule to the map container within the map instantiation.
- Adding this rule to the map container when the popup has been opened (current implementation).
- Adding this rule to the airship or whatever CSS styles.

It looks good to me adding that rule when the map is created because I agree with @jesusbotella about this behaviour is not related with Popups.

What do you think @VictorVelarde ?

VictorVelarde:
@manmorjim I haven't got a strong opinion to choose from options 1 and 2. 

If mapbox does it that way (opt 1), I guess it can be the most straight forward, better than trying to get the reference from the popup. 

I'm even thinking if just adding that to our current 'external css code' would be enough. I'm talking about this piece on the e examples:

```
      body {
        margin: 0;
        padding: 0;
      }

      #map {
        width: 100vw;
        height: 100vh;
       overflow: 'hidden';
      }
```

manmorjim:
> I'm even thinking if just adding that to our current 'external css code' would be enough. 

Yup, that also works but it implies people from CF should adapt their template by adding this rule to the container. 

In addition, future web-sdk users should add this rule to their code which is a little annoying IMHO.

VictorVelarde:
Then @manmorjim we can go after option 1, fine?

jesusbotella:
Our goal should be to properly position the popup within the screen in those edge cases. So even if the developers need to add the overflow:hidden rule to CSS (like Mapbox users do) it should not limit us when taking the decision, because they should add it regardless positioning is fixed or not.

VictorVelarde:
What do you think about option 1, @jesusbotella ?

jesusbotella:
I guess that option 1 is adding `overflow: hidden` to CSS, right? I think it is the best one, rather than getting the canvas reference which seem a little over-engineered.

Regarding the option to add `overflow:hidden` automatically to the map, I'm not sure about it. Adding it to the examples and reference should be enough, but I agree that it could be easily forgettable. So do whatever you think it's better.

manmorjim:
I've taken a look at the airship CSS and the class which has the rule `overflow: hidden` is `as-app`. So if we add this class to the map container like:

```html
...
    <main class="as-main">
      <div class="as-map-area">
        <section id="map" class="as-app"></section>
      </div>
    </main>
...
```

Then the scroll does not appear. Do you think that this should be enough?

jesusbotella:
We should not use `as-app` in there. We should add it to `.as-map-area` styles targeting the map container, or tell the developers to add it themselves.

manmorjim:
> We should not use `as-app` in there. We should add it to `.as-map-area` styles targeting the map container, or tell the developers to add it themselves.

So this should be tackled in Airship project or writing in our doc about adding the rule `overflow: hidden` to the map container, right?

Shall we close this?

manmorjim:
Finally the rule `overflow: hidden` is added to the map container at instantiation time.

VictorVelarde (approved):
🚀

PR #56, opened 2020-07-01
[Refactor API] r(e): carto.style

manmorjim:
- Removing `Style` sufix from helpers.
- Exposing `style` namespace within `viz`.


VictorVelarde (approved):
🚀  VAMOS!

PR #57, opened 2020-07-01
[Popups] Zoom jump issue (on hover)

manmorjim:
This fixes the zoom jump issue. The solutions has been adding the `onContextLoad` event which is called when deckgl executes the its event `onLoad`.

According to [the doc](https://deck.gl/docs/api-reference/core/deck#onload), this event is called once after gl context and Deck components (ViewManager, LayerManager, etc) are created.

By this way it is safe setting the view in this event:

```javascript
deckMap = carto.viz.createMap({
  basemap: basemap,
  initialViewState: camera,
  onContextLoad: () => {
    const { viewport } = deckMap.layerManager.context;
    const { longitude, latitude, zoom } = viewport.fitBounds(bounds);
    const viewState = Object.assign({}, { longitude, latitude, zoom }, camera);
    deckMap.setProps({ viewState });
  }
});
```


VictorVelarde (changes requested):
I recommended a change on the client use of web-sdk to @josemazo. Let's wait before adding more complex event handling on the map. The key point is not using initialViewState. Let's wait and decide later

VictorVelarde:
Let's wait and create an issue at deck.gl first, to explore other options

PR #59, opened 2020-07-02
[auth] Fix setCredentials

manmorjim:
This fixes the issue about widgets only works with `carto.auth.setDefaultCredentials`.

By this, widgets will use the credentials from the layer or source passed by parameter which can be different from defaults.


VictorVelarde (approved):
Do we fall back to default creds if nothing set on the Source? I suppose so, right? If we do, then go ahead!

jesusbotella (approved):

manmorjim:
> Do we fall back to default creds if nothing set on the Source? I suppose so, right? If we do, then go ahead!

That's right!

PR #60, opened 2020-07-02
fix remote dataview failing with geoJSON source

simon-contreras-deel:
https://app.clubhouse.io/cartoteam/story/89713/error-geojson-with-remote-widget-fails-with-source


VictorVelarde (approved):
🚀

PR #62, opened 2020-07-06
Local Widgets synchronization

jesusbotella:
This PR implements synchronization across widgets using local mode dataviews.

To achieve that, widgets retrieve data filtered through `getViewportFilteredFeatures` method, that returns filtered features and is capable of excluding some filters to avoid rendering only selected data within a widget.

Maybe there should not have a `getViewportFilteredFeatures` method, but a `getViewportFilteredFeatures` method that returns filtered data by default. Let me know what you think.

### DIFF
--- a/src/lib/viz/layer/Layer.ts
+++ b/src/lib/viz/layer/Layer.ts
@@ -259,19 +259,7 @@ export class Layer extends WithEvents implements StyledLayer {
     return this._deckLayer;
   }
 
-  public getViewportFeatures(properties: string[] = []) {
-    if (!this._viewportFeaturesGenerator.isReady()) {
-      throw new CartoError({
-        type: 'Layer',
-        message:
-          'Cannot retrieve viewport features because this layer has not been added to a map yet'
-      });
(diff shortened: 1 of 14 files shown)


VictorVelarde:
@jesusbotella "a getViewportFilteredFeatures method, but a getViewportFilteredFeatures" ? just a typo right?

src/lib/viz/layer/Layer.ts
```diff
@@ -270,10 +271,28 @@ export class Layer extends WithEvents implements StyledLayer {
     return this._viewportFeaturesGenerator.getFeatures(properties);
   }
 
+  public async getViewportFilteredFeatures(excludedFilters: string[] = []) {
```
VictorVelarde: Is it a bit long? Maybe we can keep the original getViewportFeatures, for familiarity with VL... IMO in both cases the concept works fine to include both the 'geometry' filter + any other filter

VictorVelarde (approved):
👌

PR #64, opened 2020-07-06
instantiation with dataview columns 

simon-contreras-deel:
Objectives:
- support dataview columns in instantiation
- fix reinstantiation regression 

Main changes:
- Fields list is now just a Set of columns
- Layer does not have the fields list, Source is responsible for that. Source is responsible to know if it needs to be reinititalized or not.
- Layer just has a new method `addSourceField` to send a new field to the source and calls to `replaceDeckGLLayer` in order to refresh everything if reinstantiation is required
- From DataViewLocal, we are using the `addSourceField`


src/lib/viz/dataview/mode/DataViewLocal.ts
```diff
@@ -14,12 +14,14 @@ export class DataViewLocal extends DataViewMode {
...
   }
 
-  public getSourceData(columns: string[] = [], options: { filterId?: string } = {}) {
+  public async getSourceData(columns: string[] = [], options: { filterId?: string } = {}) {
     if (!columns.includes(this.column)) {
       columns.push(this.column);
     }
 
     if (this.useViewport) {
+      await (this.dataSource as Layer).addSourceField(this.column);
```
VictorVelarde, line 25: Why is this conditioned just to useViewport?

src/lib/viz/layer/Layer.ts
```diff
@@ -261,11 +254,11 @@ export class Layer extends WithEvents implements StyledLayer {
   /**
    * Method to create the Deck.gl layer
    */
-  public async _createDeckGLLayer(forceInit = false) {
+  public async _createDeckGLLayer() {
```
VictorVelarde: Should we rename this method, removing the underscore, as this is public?

src/lib/viz/layer/Layer.ts
```diff
@@ -386,12 +379,12 @@ export class Layer extends WithEvents implements StyledLayer {
    * user clicks on one or more features of the layer.
    */
   public async setPopupClick(elements: PopupElement[] | string[] | null = []) {
-    this._addPopupFields(elements);
+    await this._addPopupFields(elements);
```
VictorVelarde: nice!

src/lib/viz/layer/Layer.ts
```diff
@@ -471,25 +464,22 @@ export class Layer extends WithEvents implements StyledLayer {
     return Promise.resolve();
   }
 
+  addSourceField(field: string) {
+    this._source.addField(field);
```
VictorVelarde, line 468: Should we check if the field already exists, to save the layer re-creation if it was already there?

src/lib/viz/source/Source.ts
```diff
@@ -43,27 +43,23 @@ export interface SourceProps {
...
 
     this.id = id;
-    this.isInitialized = false;
+    this.needsInitialization = true;
+    this.sourceType = 'Source';
+    this.fields = new Set();
   }
 
-  abstract async init(fields?: StatFields): Promise<boolean>;
+  abstract async init(): Promise<boolean>;
```
VictorVelarde, line 63: I see the point, it looks more clear

VictorVelarde (approved):
👌

src/lib/viz/dataview/mode/DataViewLocal.ts
simon-contreras-deel: Yeah, this is a good point and probably an smell.

In the Dataviews, `useViewport` is always `true` except in the case of GeoJSON + `remote`. So, it is going to enter here when:
- MVT + local (it is relevant here)
- GeoJSON + local (it does not matter)
- GeoJSON + remote filtered (it does not matter)

Do you think I should move it out of the `if` in order to avoid all of this?

src/lib/viz/layer/Layer.ts
simon-contreras-deel: Yes, we need to follow always the same convention (I prefer not to use underscore with TS). In this case, it is in fact a private method, so, I will change it

src/lib/viz/layer/Layer.ts
simon-contreras-deel: The logic about reinitialized is managed by Source class where we have:
```
addField(field: string) {
  const { size } = this.fields;

  this.fields.add(field);

  if (size < this.fields.size) {
    this.needsInitialization = true;
  }
}
```

So the fields are a Set in order to avoid duplicates and in case a new element is added, `needsInitialization` is set to `true` and in the following `replaceDeckGLLayer` the source will be reinitialized

src/lib/viz/dataview/mode/DataViewLocal.ts
VictorVelarde: I think this should be tackled in the second PR that @manmorjim will do (internal refactor), after the current WIP in the public API change. So I will merge this one, and discuss in the context of the new 'mode parameters'

src/lib/viz/layer/Layer.ts
VictorVelarde: Perfect, I didn't notice the Set.

VictorVelarde:
From my perspective, this has a good approach. Please try to merge it in after fixing the conflicts. When in doubt, Manu is a nice help as he's working also on this area

PR #65, opened 2020-07-06
[Popup & Interactivity] Add and improve unit tests for popups

manmorjim:
- Update element popup types according to the API
- Added unit tests for popups


VictorVelarde (approved):
👏  Go go go!

PR #66, opened 2020-07-07
Easy access isOpened property in Popup.

sclebal:
I've added a method in Popup to check if a popup is already opened or not. It's very useful.

### DIFF
--- a/src/lib/viz/popups/Popup.ts
+++ b/src/lib/viz/popups/Popup.ts
@@ -25,14 +25,14 @@ export class Popup {
   private _deckInstance: Deck | undefined;
   private _container: HTMLElement;
   private _parentElement: HTMLElement | undefined;
-  private _isOpened: boolean;
+  private _isOpen: boolean;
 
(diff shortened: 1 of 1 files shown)


manmorjim (commented):
LGTM :ok_hand: 

Now Popup tests are available in `develop` it would be nice adding some tests for this method :blush:

VictorVelarde:
I was going to suggest renaming the internal '_isOpened' variable to '_isOpen' and turn the function into a property, like this:

```
  /**
   * Whether the popup is open or not
   */
  public get isOpen() {
    return this._isOpen;
  }
```

sclebal:
@VictorVelarde changed and updated with the fix of this task https://app.clubhouse.io/cartoteam/story/87541/popup-interactivity-glitch-when-cursor-hover-the-popup. It should work. Maybe I've should opened a different PR for the fix 🤦 sorry, the rush is not good.

VictorVelarde (approved):
LGTM. Let's merge this!

PR #67, opened 2020-07-07
Added an option property to addTo method to allow customise the posit…

sclebal:
…ion where the layer is added.

POC


VictorVelarde:
Thx a lot Sergio, I have just add some tests and a bit of doc, and I think we're ready to go

sclebal:
@VictorVelarde very good improvements. The only thing I miss is to create an interface for options `{ beforeLayerId?: string; afterLayerId?: string }` instead of repeating it.

VictorVelarde:
Good point. Here it is!

VictorVelarde (approved):

PR #68, opened 2020-07-07
Remove d3 format option from PopUps

VictorVelarde:
This is a small PR to remove the dependency on **d3.format** for the library. 

Main reasons:
- I think it doesn't provide much value and it adds a dependency.
- you can get exactly the same results importing it externally, from your project. 

So instead of
   ```
      {
         attr: 'pop',
         title: 'Population D3',
         format: '~s'
    }
  ```

the user would use the general option to pass any function

   ```
      {
         attr: 'pop',
         title: 'Population D3',
         format: d3.format('~s')
    }
  ```

As a complement:
- This will also work in the future when we consider date fields (the user might use externally for example `moment.js`) 
- We save some KBs (just a few).


VictorVelarde:
It looks like we have some failing test, related to histogram @jesusbotella

jesusbotella:
I see that tests are passing right now. Did you re-run them?

VictorVelarde:
GITHUB issue? I'm not sure why, but it looks like here we have some comments that belongs to another PR (at least in my browser I get some un-related comments about 'histogram', that don't belong to this).

jesusbotella:
Yes, not sure. I saw that histogram comment as well 😅 , but tests are passing.

VictorVelarde:
I ran them again, yes!

VictorVelarde:
.... ok, so coming back to this issue.... Please @manmorjim have a look at it

src/lib/viz/popups/Popup.ts
```diff
@@ -303,16 +294,8 @@ export interface PopupElement {
    * Title for this element.
    */
   title?: string | null;
-
-  /**
-   * d3 format for the value of this attribute.
-   */
-  format?: string | FormatFunction;
```
manmorjim, line 310: we should keep the `format` option for functions right?

```javascript
format?: (value: any) => any;
```

src/lib/viz/popups/Popup.ts
```diff
@@ -267,19 +266,11 @@ function generatePopupContent(elements: any, features: Record<string, any>[]): s
...
-              formatter = d3Format(format);
-            } catch (err) {
-              throw new CartoPopupError(
-                `The format '${format}' is not a recognized D3 format`,
-                popupErrorTypes.FORMAT_INVALID
-              );
-            }
-
-            elementValue = formatter(elementValue);
+          } else if (format && typeof format !== 'function') {
```
manmorjim: the second condition is not necessary

```javascript
...
} else if (format) {
...
```

manmorjim (changes requested):
Everything looks great to me :ok_hand: 

Just a small request to keep the `format` options for custom functions.

src/lib/viz/popups/Popup.ts
VictorVelarde: I thought it was just related to d3... oka

src/lib/viz/popups/Popup.ts
VictorVelarde: for sure! thx

VictorVelarde:
Rebasing to get changes from develop...

src/lib/viz/popups/Popup.ts
manmorjim: Yup sorry, the comment was outdated and wrong :sweat_smile:

manmorjim (approved):
:rocket:

PR #69, opened 2020-07-07
Bug/ch87541/popup interactivity glitch when cursor hover

sclebal:
This PR seems to fix 87541 task.

You must merge #66 first.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 - Remove d3.format option from `carto.viz.Popup`
 
 ### Fixed
-- 
+- Fix popup interactivity glitch on hover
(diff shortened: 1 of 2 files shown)


manmorjim (approved):
LGTM :ok_hand: 

I think we can check the condition inside the method `_render` but this is just a small comment.

```javascript
private _render() {
  if (this._isOpen && this._coordinates && this.getContent().trim().length > 0 && this._deckInstance) {
    const pixels = coordinates2pixels(this._coordinates, this._deckInstance);
...
}
```

sclebal:
@manmorjim I think it's ok. Uploaded 👌

VictorVelarde:
LGTM. Let's just add a small note in the CHANGELOG (that we didn't have before), with a small description of the fix, and we're good to go

VictorVelarde (approved):

PR #72, opened 2020-07-08
[DataView] Refactor API - Public API

manmorjim:
## Public API proposal

```javascript
const dataview = new carto.viz.dataview.Category(countriesLayer, 'continent', {
  operation: 'count',
  mode: 'precise',
  filters: {
    foo_column: {in: ['value1', 'value2']},
    bar_column: {in: ['value3']}
  },
  spatialFilter: 'viewport'
});

dataview.setFilters({
  foo_column: {in: ['value1', 'value2']},
  my_column: {in: ['others']},
});

dataview.setSpatialFilter({
  bbox: [64.3, -7.2, 33.12, -29.56]
});

new carto.viz.widget.Category('as-category-widget', dataview);
```

### Strategies

**MVT all features (remote only)**
```javascript
const dataview = new carto.viz.dataview.Category(datasetLayer, 'continent', {
  operation: 'count',
  // mode: 'precise' // ignored
});
```

**MVT viewport features (local)**
```javascript
const dataview = new carto.viz.dataview.Category(datasetLayer, 'continent', {
  operation: 'count',
  // mode: 'fast', // default
  spatialFilter: 'viewport'
});
```

**MVT viewport features (remote)**
```javascript
const dataview = new carto.viz.dataview.Category(datasetLayer, 'continent', {
  operation: 'count',
  mode: 'precise',
  spatialFilter: 'viewport'
});
```

**GeoJSON all features (local)**

```javascript
const dataview = new carto.viz.dataview.Category(geojsonLayer, 'continent', {
  operation: 'count',
  // mode: 'fast' | 'precise' // ignored
});
```

**GeoJSON viewport features (local)**

```javascript
const dataview = new carto.viz.dataview.Category(geojsonLayer, 'continent', {
  operation: 'count',
  // mode: 'fast' | 'precise', // ignored
  spatialFilter: 'viewport'
});
```


VictorVelarde:
Everything looks fine to me, but the GeoJSON difference drives me nuts. I think we must solve it, it's not clear. We should have the same behaviour, but I'm still trying to figure how

manmorjim:
> Everything looks fine to me, but the GeoJSON difference drives me nuts. I think we must solve it, it's not clear. We should have the same behaviour, but I'm still trying to figure how

Yes, in fact, for GeoJSON layers `remote` mode is equivalent to `local` due to its nature. However, no use cases come to my mind where the user would like getting just the viewport features. If we have all the features loaded in the browser (which is one of the benefits of GeoJSON layers) why would we want to filter them by the viewport? so I think we should keep the same behavior as MVT and make equivalent both `local` modes: with viewport filter and without it

manmorjim:
After a quick discussion about the differences between the GeoJSON layers and MVT layers behaviors for the modes, what do you think about the following proposal?

For GeoJSON layers the mode `remote` and `local` are equivalents. With this, we have:


```javascript
const dataview = new carto.viz.dataview.Category(geojsonLayer, 'continent', {
  operation: 'count',
  mode: 'remote'
});
```
:top: This will provide **all the GeoJSON features** (similar to MVT). 


```javascript
const dataview = new carto.viz.dataview.Category(geojsonLayer, 'continent', {
  operation: 'count',
  mode: 'remote',
  spatialFilter: {
    viewport: true
  }
});
```
:top: This will provide **the GeoJSON features which are loaded in the viewport** (similar to MVT). 


```javascript
const dataview = new carto.viz.dataview.Category(geojsonLayer, 'continent', {
  operation: 'count',
  mode: 'local'
});
```
:top: This will provide **all the GeoJSON features** (similar to `remote`)

```javascript
const dataview = new carto.viz.dataview.Category(geojsonLayer, 'continent', {
  operation: 'count',
  mode: 'local',
  spatialFilter: {
    viewport: true
  }
});
```
:top: This will provide **the GeoJSON features which are loaded in the viewport** (similar to MVT).

examples/dataview/category-data-view-all-modes.html
```diff
@@ -94,9 +94,9 @@
       );
     const errorContainer = document.getElementById('errorContainer');
 
-    let categoryRemote;
-    let categoryRemoteFiltered;
-    let categoryLocal;
+    let categoryGlobal;
+    let categoryViewportPrecise;
```
VictorVelarde, line 98: I think we should just keep global & viewport for public examples (not entering into the 'fast' vs 'precise' modes at all. I would like to ask you if you please could remove that duality. 

And if you see something relevant for us in this example, move it to a debug example. Could you please?

examples/dataview/formula-data-view-all-modes.html
```diff
@@ -161,9 +160,9 @@
       const formulaLocalElement = document.getElementById('formula-local');
       const errorContainer = document.getElementById('errorContainer');
 
-      let formulaRemote;
-      let formulaRemoteFiltered;
-      let formulaLocal;
+      let formulaGlobal;
+      let formulaViewportPrecise;
```
VictorVelarde, line 164: Same here, II would just call the example 'formulaDataView' and remove the viewport variants, leaving just the default

VictorVelarde: I see now we have another one with that name. I guess we could mix them both together, having 'global' and 'viewport'

examples/widgets/histogram.html
```diff
@@ -119,7 +119,8 @@ <h4 class="as-subheader as-mb--12">
             bins: 10,
             start: 1000000,
             end: 127078679,
-            mode: 'REMOTE_FILTERED'
+            spatialFilter: 'viewport',
+            mode: 'precise'
```
VictorVelarde: Probably the same here --> let's have a debug example with variants 'precise' / 'fast', but leave just the defaults for public examples

src/lib/viz/dataview/category/CategoryDataView.ts
```diff
@@ -11,44 +9,32 @@ import {
...
-        const credentials = getCredentialsFrom(dataSource);
-        dataView = new DataViewRemote(dataSource, column, credentials);
-        break;
-      }
-
-      case DataViewCalculation.REMOTE_FILTERED:
+    const geoJSONSource = isGeoJSONSource(dataSource);
+
+    if (this.mode === DataViewCalculation.FAST || geoJSONSource) {
+      const useViewport = !(this.mode === DataViewCalculation.PRECISE && geoJSONSource);
```
VictorVelarde: Can't we get 'viewport' use from spatialFilter?

src/lib/viz/dataview/formula/FormulaDataView.ts
```diff
@@ -1,51 +1,37 @@
...
-
-        const credentials = getCredentialsFrom(dataSource);
-        dataView = new DataViewRemote(dataSource as Source, column, credentials);
-        break;
-      }
-
-      case DataViewCalculation.REMOTE_FILTERED:
+    const geoJSONSource = isGeoJSONSource(dataSource);
+
+    if (this.mode === DataViewCalculation.FAST || geoJSONSource) {
```
VictorVelarde: Aren't these lines the same here than in CategoryDataView? If that's the case, we can push them out to another class, up in the hierarchy... a getDataViewFor(source, calculationMode), that hides that 'table' withe the logic

VictorVelarde (approved):
Public API looks great. Take my comments as suggestions, that could go to #77

src/lib/viz/dataview/category/CategoryDataView.ts
manmorjim: This was already addressed in #77

src/lib/viz/dataview/formula/FormulaDataView.ts
manmorjim: nice suggestion!

PR #63, opened 2020-07-06
Remote Widgets synchronization

jesusbotella:
This PR implements remote widgets synchronization, same as #62 but for widgets using remote and remote filtered dataviews.

PR #73, opened 2020-07-08
Fix filters bug in release

jesusbotella:
This PR fixes a bug introduced in the alpha release.


VictorVelarde:
@jesusbotella Can we add a little failing test and then the fix? Also a small note in the CHANGELOG.md would be great!

jesusbotella:
This should not be required, because of a misunderstanding on #63. I'll add the tests there.

PR #77, opened 2020-07-13
[DataView] Refactor API - Internal refactor

manmorjim:
### Internal changes

- Removed flag `useViewport` from constructor. Set when the filter `viewport` is applied.
- Implemented the method `setSpatialFilter` for `DataViewLocal`.
- Added the method `getFeatures` in layers which returns all the features. For MVT layers it will return just viewport features.
- Added the method `getFeatures` in sources.


src/lib/viz/dataview/DataView.ts
```diff
@@ -42,11 +42,9 @@ export abstract class DataView<T> extends WithEvents {
 
   protected createDataViewMode(column: string, options: DataViewOptions): DataViewMode {
     let dataViewMode;
-    const geoJSONSource = isGeoJSONSource(this.dataSource);
 
-    if (this.mode === DataViewCalculation.FAST || geoJSONSource) {
-      const useViewport = !(this.mode === DataViewCalculation.PRECISE && geoJSONSource);
-      dataViewMode = new DataViewLocal(this.dataSource as Layer, column, useViewport);
+    if (this.mode === DataViewCalculation.FAST || isGeoJSONSource(this.dataSource)) {
+      dataViewMode = new DataViewLocal(this.dataSource as Layer, column);
```
VictorVelarde, line 47: Nice removal!

src/lib/viz/dataview/mode/DataViewLocal.ts
```diff
@@ -60,9 +55,8 @@ export class DataViewLocal extends DataViewMode {
     }
   }
 
-  // eslint-disable-next-line @typescript-eslint/no-unused-vars, class-methods-use-this
-  public setSpatialFilter(_spatialFilter: SpatialFilters) {
-    return undefined;
+  public setSpatialFilter(spatialFilter: SpatialFilters) {
+    this.useViewport = spatialFilter === 'viewport';
```
VictorVelarde, line 59: I'm ok with this, but I guess that in the future we might need the whole spatialFilter stored

src/lib/viz/layer/Layer.ts
```diff
@@ -299,6 +299,14 @@ export class Layer extends WithEvents implements StyledLayer {
     return features;
   }
 
+  public async getFeatures(excludedFilters: string[] = []) {
```
VictorVelarde, line 302: @manmorjim I'm not sure if I like the idea of having a common & public `getFeatures` here, with 2 different meanings... And we also are going to keep the `getViewportFeatures`, so it's also difficult to explain for the final user.

I think we should probably remove its use here in the Layer, and keep it in the `Source` where it belongs (IMO this is more a shortcut for our internal use than a clear feature for the user). We should review its external use (I believe it is just in the Dataview internals...) to see if we can avoid it.

I don't want to block this PR, cause I think other changes are good to go, so if you want to discuss more about this, we would merge this and then we will have a bit more of discussion

VictorVelarde (approved):
OK. I have doubts about the 'getFeatures' on Layer, but feel free to merge and then discuss

src/lib/viz/layer/Layer.ts
manmorjim: Yup, I'd like discussing more about you suggested because I agree with you about it is not a clear from a user point of view having two different methods `getViewportFeatures` and `getFeatures`.
But in the internal use is quite weird the casuistry:
- Layers with `SQLSource` or `DatasetSource` use `getViewportFeatures`.
- Rest of layers use `getFeatures` from their sources.

As you said, let's merge this and having a bit more of discussion.

VictorVelarde: Agree. Let's merge. 

My 2 cts for the next discussion: right now we can't get the 'whole features' (that's what I understand with 'getFeatures') for tiled layers (maybe in the feature, with a paginated query to SQL, for example).... so...

I see no issue in having only what we can support always: the getViewportFeatures, and leave the 'getFeatures' to those sources which can support it

manmorjim: Ticket created in CH for the discussion: https://app.clubhouse.io/cartoteam/story/93117/discussion-merge-methods-getfeature-and-getviewportfeatures

PR #78, opened 2020-07-14
FIX Layer with empty data

manmorjim:
CH: https://app.clubhouse.io/cartoteam/story/90394


src/lib/viz/source/SQLSource.ts
```diff
@@ -226,11 +227,14 @@ export class SQLSource extends Source {
...
   private extractMetadataFrom(mapInstance: MapInstance) {
     const { stats } = mapInstance.metadata.layers[0].meta;
-    const geometryType = parseGeometryType(stats.geometryType);
-    const fieldStats = this.getCompleteFieldStats(stats);
+    let geometryType: GeometryType | undefined;
 
-    const metadata = { geometryType, stats: fieldStats };
+    if (stats.geometryType) {
+      geometryType = parseGeometryType(stats.geometryType);
+    }
```
VictorVelarde, line 238: Should we at least add a console.warn about empty source? Or maybe even have a source.isEmpty property based on this value?

src/lib/viz/style/helpers/color-bins-style.ts
```diff
@@ -50,6 +50,11 @@ export function colorBinsStyle(
 ) {
   const evalFN = (layer: StyledLayer) => {
     const meta = layer.source.getMetadata();
+
+    if (!meta.geometryType) {
```
VictorVelarde: I would better go after a `layer.source.isEmpty` prop, hiding this... what do you think?

src/lib/viz/source/SQLSource.ts
```diff
@@ -226,11 +227,14 @@ export class SQLSource extends Source {
 
   private extractMetadataFrom(mapInstance: MapInstance) {
     const { stats } = mapInstance.metadata.layers[0].meta;
-    const geometryType = parseGeometryType(stats.geometryType);
-    const fieldStats = this.getCompleteFieldStats(stats);
+    let geometryType: GeometryType | undefined;
 
-    const metadata = { geometryType, stats: fieldStats };
+    if (stats.geometryType) {
```
simon-contreras-deel, line 236: Maybe it easier to do something like:

```
if (!stats.geometryType) {
  return { 'Point', stats: [] }
}
```

And even better, if you encapsulate the check `!stats.geometryType` in a method called `hasData`.

I would expect you don't need to check if `geometryType` exists and it is not going to fail for that reason in other places

VictorVelarde (approved):
It looks good. I just added a couple of small suggestions on how the source is asked about 'isEmpty'

src/lib/viz/source/SQLSource.ts
VictorVelarde: Same thoughts here around 'hasData' or 'isEmpty' method

manmorjim:
Not sure if I understood your suggestions:

 I've added a method `isEmpty` in `Source` class which is overwritten in the class `SQLSource`. The `isEmpty` method of this class returns `true` if its metadata attribute does not have the `geometryType` field. Then, style helpers check if the source is empty or not using this method.

src/lib/viz/source/SQLSource.ts
```diff
@@ -214,6 +215,10 @@ export class SQLSource extends Source {
     }
   }
 
+  public isEmpty() {
```
VictorVelarde, line 218: Nice, I think this is useful. Although I see the default method in the generic Source, I'm missing it in the GeoJSON one. I think we should try to keep the Sources as aligned as possible. What do you think? In that case I would return something like "features.length === 0"…

src/lib/viz/source/Source.ts
```diff
@@ -60,6 +60,11 @@ export abstract class Source extends WithEvents {
     this.registerAvailableEvents(['filterChange']);
   }
 
+  // eslint-disable-next-line class-methods-use-this
+  public isEmpty() {
```
VictorVelarde: Why not an abstract method? I find strange to see an implementation in the interface...

VictorVelarde:
@manmorjim Some small comments / suggestions. Feel free to merge after considering them

src/lib/viz/source/Source.ts
manmorjim: I wanted to add a _default_ implementation to avoid having to add the method in every source. But I reconsidered your suggestion and I agree that every source should implement its own method.

PR #79, opened 2020-07-14
Formula DataView Aggregated Calculations

jesusbotella:
This PR implements aggregated calculations for Formula DataView.

So, I have added a new method to Layer class, `addAggregationOptions`, to add aggregation configuration to source (aggregated columns, and dimensions).

Once we get that data aggregated within feature properties, we check if the aggregated column is present to perform calculations.


src/lib/viz/dataview/formula/FormulaDataViewImpl.ts
```diff
@@ -9,14 +9,29 @@ export class FormulaDataViewImpl extends DataViewImpl<FormulaDataViewData> {
...
+          numeric: [
+            {
+              column: this.column,
+              operations: [this.operation]
+            }
+          ]
+        }
+      });
+
+      const aggregatedColumnName = `${this.operation}__${this.column}`;
```
VictorVelarde: how does the percentile 'name' fit into this? And it's not easy to get collisions, but probably not so rare, so maybe we should add a preffix to avoid them, like '_cdb_' + operation...  We could move this 'build-the-name-for-column' into a little external function... what do you think?

VictorVelarde (approved):
LGTM. It would be nice to have some feedback from @oleurud in the next days, regarding to any possible simplification to come

src/lib/viz/dataview/formula/FormulaDataViewImpl.ts
VictorVelarde: How about the '_cdb' preffix... did you consider something like that?

src/lib/viz/dataview/formula/FormulaDataViewImpl.ts
```diff
@@ -23,10 +23,13 @@ export class FormulaDataViewImpl extends DataViewImpl<FormulaDataViewData> {
...
-      // eslint-disable-next-line @typescript-eslint/no-explicit-any
-      const values = features.map(
-        (feature: Record<string, unknown>) => feature[aggregatedColumnName] || feature[columnName]
-      );
+      const values = features.map((feature: Record<string, unknown>) => {
+        const aggregatedValue = feature[aggregatedColumnName] as number;
+        const value = feature[columnName] as number;
+        const clusterCount = feature._cdb_feature_count as number;
+
+        return (aggregatedValue || value) * clusterCount;
```
VictorVelarde: @jesusbotella Are you sure about this generic calculation approach for all?, Does it fit 'MIN, MAX, COUNT...'? Wouldn't we get a wrong MAX if we multiply by clusterCount?

We have the VL implementation to check at https://github.com/CartoDB/carto-vl/tree/v1.4.4/src/renderer/viz/expressions/aggregation/viewport, if it's required

VictorVelarde (commented):
I have some doubts about the clusterCount in the formula calculations...

src/lib/data/operations/aggregation/feature/feature-aggregation.spec.ts
```diff
@@ -0,0 +1,59 @@
...
+      expect(() => aggregateFeatures(values, 'unknownAggregation' as AggregationType)).toThrow(
+        new CartoError({
+          type: '[DataView]',
+          message: '"unknownAggregation" feature aggregation type not implemented'
+        })
+      );
+    });
+  });
+
+  describe('aggregate', () => {
```
VictorVelarde, line 17: Cool tests. This is what we need to test our assertions 👍

src/lib/data/operations/aggregation/feature/feature-aggregation.ts
```diff
@@ -0,0 +1,60 @@
...
+}
+
+// eslint-disable-next-line @typescript-eslint/ban-types
+const aggregationFunctions: Record<AggregationType, Function> = {
+  [AggregationType.COUNT](aggregatedFeatures: AggregatedFeatureProperties[]) {
+    return aggregatedFeatures.reduce((total, feature) => total + feature.featureCount, 0);
+  },
+
+  [AggregationType.AVG](aggregatedFeatures: AggregatedFeatureProperties[]) {
+    return aggregationFunctions.sum(aggregatedFeatures) / aggregatedFeatures.length;
```
VictorVelarde: Shouldn't this average be ponderated by the count? Can we reuse the aggreg. count and aggr. sum?

jesusbotella: Yes, I did that at the beginning but then I had a conflict about AVGs and forgot to change it back 😁

VictorVelarde (commented):
Ok, let's move on and deal with percentile in a next PR

PR #80, opened 2020-07-14
[Popup] Adjust width to the content

manmorjim:
CH: https://app.clubhouse.io/cartoteam/story/90447

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,8 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 - New `BQSource` for public datasets ([#122](https://github.com/CartoDB/web-sdk/pull/122/))
 - New `onLoad` option to `carto.viz.createMap` ([#121](https://github.com/CartoDB/web-sdk/pull/121/))
 - Better performance in Layer, with a debounced approach to DATA_CHANGED event ([#130](https://github.com/CartoDB/web-sdk/pull/130))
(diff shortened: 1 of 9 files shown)


src/lib/viz/popups/Popup.ts
```diff
@@ -110,6 +110,13 @@ export class Popup {
   public open() {
     if (this._parentElement && !this._isOpen) {
       this._parentElement.appendChild(this._container);
+
+      // adjust width to content
+      const contentWrapper = this._container.querySelector('div.as-infowindow') as HTMLElement;
+
+      if (contentWrapper) {
+        contentWrapper.style.width = 'calc(100% + 17px)';
```
VictorVelarde: why this magic number of 17px? maybe a little comment?

VictorVelarde:
Probably you already know... but it looks like the 'first-time' the with is not adjusted yet, for example in https://web-sdk-git-feature-ch90447popup-adjusting-the-po-afdce0.carto-frontend.now.sh/examples/interactivity/add-popups-onhover.html

VictorVelarde:
Blocked till we decide on the approach to fix the issue

VictorVelarde:
Converting this into an optional flag for Popup: { autoWidth: true },. that can be set from the Layer (setPopupClick / setPopupHover)

image

🔝 autowidth: true

... up to a maximum of 400px (set on Airship):
image

VictorVelarde (approved):
👍

PR #82, opened 2020-07-15
improve error message using an invalid column for size category helper

simon-contreras-deel:
When using an invalid column for size category helper, the error was:

`The featureProperty has not categories`

I am changing it to:

`Current dataset has not categories for 'your_column_used'`


manmorjim (approved):
nice :ok_hand:

PR #84, opened 2020-07-15
improve data events

simon-contreras-deel:
fix https://app.clubhouse.io/cartoteam/story/92672/map-refine-map-related-data-events

The idea is to create a state object which is responsible of:
- getting event inputs: from deck instance (zoom, pan, rotate, tiles map loaded and onAfterRender)
- emitting 2 new events: 'layer:data:ready' and layer:data:changed' when is required 

The objectives are:
- solving the problems
- manage the events in one place

By now it is also emitting the previous one `viewportLoad` which is used by dataviews, but the idea is to change to the new ones.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 - Add `beforeLayerId` and `afterLayerId` options to `carto.viz.Layer.addTo` method, to customize layer position
 - Widget synchronization when using remote dataviews ([#76](https://github.com/CartoDB/web-sdk/pull/76/))
 - `icon` style helper 
-- Adding new data state in Layer ([#84](https://github.com/CartoDB/web-sdk/pull/84)) for sending data events when data is ready or updated
+- New data events in Layer: `layerDataReady` and `layerDataChanged` ([#84](https://github.com/CartoDB/web-sdk/pull/84))
 
 ### Changed
 - Remove d3.format option from `carto.viz.Popup`
@@ -19,5 +19,3 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 
 ## [1.0.0-alpha.0] 2020-07-06
 - First release
-
-
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 - Add `beforeLayerId` and `afterLayerId` options to `carto.viz.Layer.addTo` method, to customize layer position
 - Widget synchronization when using remote dataviews ([#76](https://github.com/CartoDB/web-sdk/pull/76/))
 - `icon` style helper 
+- Adding new data state in Layer ([#84](https://github.com/CartoDB/web-sdk/pull/84)) for sending data events when data is ready or updated
 
 ### Changed
 - Remove d3.format option from `carto.viz.Popup`
--- a/src/lib/viz/layer/Layer.ts
+++ b/src/lib/viz/layer/Layer.ts
@@ -72,6 +72,7 @@ export class Layer extends WithEvents implements StyledLayer {
     this.registerAvailableEvents([
       DATA_READY_EVENT,
       DATA_CHANGED_EVENT,
+      'viewportLoad',
       'filterChange',
       InteractivityEventType.CLICK.toString(),
       InteractivityEventType.HOVER.toString()
@@ -83,7 +84,7 @@ export class Layer extends WithEvents implements StyledLayer {
     };
 
     this._interactivity = this._buildInteractivity(options);
-    this.buildDataState();
+    this.dataState = buildDataState();
   }
 
   getMapInstance(): Deck {
@@ -190,7 +191,7 @@ export class Layer extends WithEvents implements StyledLayer {
         }
       },
       onAfterRender: () => {
-        this.sendDataEvent();
+        this.sendDataEvent('onAfterRender');
       }
     });
 
@@ -324,7 +325,7 @@ export class Layer extends WithEvents implements StyledLayer {
           styleProperties.onViewportLoad(...args);
         }
 
-        this.emit('viewportLoad');
+        this.sendDataEvent('onViewportLoad');
       },
       onClick: this._interactivity.onClick.bind(this._interactivity),
       onHover: this._interactivity.onHover.bind(this._interactivity)
@@ -509,13 +510,6 @@ export class Layer extends WithEvents implements StyledLayer {
     }
   }
 
-  private buildDataState() {
-    this.dataState.isFirstTime = false;
-    this.dataState.isPanning = false;
-    this.dataState.isZooming = false;
-    this.dataState.isRotating = false;
-  }
-
   private saveDataState(
     isPanning: boolean,
     isZooming: boolean,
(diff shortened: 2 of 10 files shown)


src/lib/viz/layer/Layer.ts
```diff
@@ -21,6 +22,16 @@ import { FunctionFilterApplicator } from '../filters/FunctionFilterApplicator';
...
 import { basicStyle } from '../style/helpers/basic-style';
 
+const DATA_READY_EVENT = 'layer:data:ready';
+const DATA_CHANGED_EVENT = 'layer:data:changed';
+
+export interface LayerDataState {
+  isFirstTime: boolean;
+  isPanning: boolean;
+  isZooming: boolean;
+  isRotating: boolean;
```
VictorVelarde: I'm not sure that we want to distinguish these 'subtypes'. We might miss some (for example what happens if we change the pitch?) but in any case, we always treat them together, so I would prefer a common state, named more generically, like 'isViewChanging' or something similar...

src/lib/viz/layer/Layer.ts
```diff
@@ -59,6 +70,8 @@ export class Layer extends WithEvents implements StyledLayer {
     this._style = buildStyle(style);
 
     this.registerAvailableEvents([
+      DATA_READY_EVENT,
+      DATA_CHANGED_EVENT,
       'viewportLoad',
```
VictorVelarde: I like the idea of CONSTANTS for events here, indeed we could also add them for the others

src/lib/viz/layer/Layer.ts
```diff
@@ -21,6 +22,16 @@ import { FunctionFilterApplicator } from '../filters/FunctionFilterApplicator';
 import { ColumnFilters } from '../filters/types';
 import { basicStyle } from '../style/helpers/basic-style';
 
+const DATA_READY_EVENT = 'layer:data:ready';
```
VictorVelarde: deckgl uses 'camelCase', so how about 'layerDataReady'?

src/lib/viz/layer/Layer.ts
```diff
@@ -499,6 +502,57 @@ export class Layer extends WithEvents implements StyledLayer {
...
+      return false;
+    }
+
+    this.dataState.isPanning = isPanning;
+    this.dataState.isZooming = isZooming;
+    this.dataState.isRotating = isRotating;
+
+    const isGeoJsonLayer = this._source.sourceType === 'GeoJSONSource';
+
+    if ((isPanning || isZooming) && isGeoJsonLayer) {
```
VictorVelarde: Why are we missing the 'isRotating' here?

src/lib/viz/layer/Layer.ts
VictorVelarde: Just to add some other angle in case it is useful: this is VL approach: https://github.com/CartoDB/carto-vl/blob/v1.4.4/src/Layer.js#L19 simplified to 3 states: initial / updating / idle

src/lib/viz/layer/Layer.ts
```diff
@@ -46,7 +57,7 @@ export class Layer extends WithEvents implements StyledLayer {
   private filtersCollection = new FiltersCollection<ColumnFilters, FunctionFilterApplicator>(
     FunctionFilterApplicator
   );
-  private callToViewportLoad = false;
+  private dataState: LayerDataState;
```
VictorVelarde: I like the idea of having a clear state, 👍 1 for this

src/lib/viz/layer/Layer.ts
```diff
@@ -315,7 +318,7 @@ export class Layer extends WithEvents implements StyledLayer {
           styleProperties.onViewportLoad(...args);
         }
 
-        this.emit('viewportLoad');
+        this.sendDataEvent('onViewportLoad');
```
VictorVelarde, line 327: Are we going to keep this?

src/lib/viz/layer/Layer.ts
```diff
@@ -21,6 +22,16 @@ import { FunctionFilterApplicator } from '../filters/FunctionFilterApplicator';
 import { ColumnFilters } from '../filters/types';
 import { basicStyle } from '../style/helpers/basic-style';
 
+const DATA_READY_EVENT = 'layer:data:ready';
+const DATA_CHANGED_EVENT = 'layer:data:changed';
+
+export interface LayerDataState {
+  isFirstTime: boolean;
```
VictorVelarde: Does 'isFirstTime' correspond to 'no-data-yet', so it's the first state of the 'Layer'?

src/lib/viz/layer/Layer.ts
```diff
@@ -499,6 +502,57 @@ export class Layer extends WithEvents implements StyledLayer {
...
+    this.dataState.isRotating = isRotating;
+
+    const isGeoJsonLayer = this._source.sourceType === 'GeoJSONSource';
+
+    if ((isPanning || isZooming) && isGeoJsonLayer) {
+      const viewport = new WebMercatorViewport(viewState);
+      this._viewportFeaturesGenerator.setViewport(viewport);
+    }
+
+    return true;
```
VictorVelarde: do we use this as confirmation anywhere?

src/lib/viz/layer/Layer.ts
```diff
@@ -576,3 +631,12 @@ interface LayerPosition {
   beforeLayerId?: string;
   afterLayerId?: string;
 }
+
+function buildInitialDataState() {
+  return {
+    isFirstTime: true,
```
VictorVelarde: If we join 'isPanning' / 'isZooming'... we might consider using an enum instead of a set of flags, so only 1 state is possible at the same time...

src/lib/viz/layer/Layer.ts
simon-contreras-deel: I think they could be interesting in the future knowing the reason, but know in the current implementation it does not matter

src/lib/viz/layer/Layer.ts
simon-contreras-deel: Yes, we need it

src/lib/viz/layer/Layer.ts
simon-contreras-deel: the idea behind it is: "no events send yet" or being more specific, "ready event not sent yet". I can change it to `isReady` if it is more clear

src/lib/viz/layer/Layer.ts
simon-contreras-deel: no

src/lib/viz/layer/Layer.ts
simon-contreras-deel: Yep, I think it is possible to do so

src/lib/viz/layer/Layer.ts
simon-contreras-deel: Ok, I have not a real preference

src/lib/viz/layer/Layer.ts
simon-contreras-deel: Changed to use 1 unique state, so this one does not applies

src/lib/viz/layer/Layer.ts
simon-contreras-deel: Done in 2 steps if you want to follow the code

src/lib/viz/layer/Layer.ts
simon-contreras-deel: Changed to use 1 unique state, so this one does not applies

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 - Add `beforeLayerId` and `afterLayerId` options to `carto.viz.Layer.addTo` method, to customize layer position
 - Widget synchronization when using remote dataviews ([#76](https://github.com/CartoDB/web-sdk/pull/76/))
 - `icon` style helper 
+- Adding new data state in Layer ([#84](https://github.com/CartoDB/web-sdk/pull/84)) for sending data events when data is ready or updated
```
VictorVelarde: The state of Layer is relevant, but specially for us, not so much externally for the end user. I would add a note instead more on the line of 'New data events in Layer: XXX an YYY'... what do you think?

We didn't add the PR in other notes, but it's nice, we can try to add it always.

src/lib/viz/layer/Layer.ts
```diff
@@ -94,6 +107,7 @@ export class Layer extends WithEvents implements StyledLayer {
 
     if (this._deckLayer) {
       await this.replaceDeckGLLayer();
+      this.dataState = DATA_STATES.STARTING;
```
VictorVelarde, line 109: This is applied to `setSource`. Would it be also necessary with `setStyle`?

simon-contreras-deel: No, it would not because data does not change with setStyle

VictorVelarde (approved):
LGTM. Small suggestion of renaming 'layerDataReady' --> 'dataReady'

PR #85, opened 2020-07-15
[DataView] Aggregated calculations - CategoryDataView

manmorjim:
### Main changes

- Added the type `GetDataOptions` for the options parameter in `getLocalData` and `getRemoteData`.
- Add `column` attribute of DataViews to the layer and source for aggregations.


src/lib/viz/dataview/mode/DataViewMode.ts
```diff
@@ -14,6 +14,12 @@ export abstract class DataViewMode extends WithEvents {
 
     this.column = column;
     this.dataSource = dataSource;
+
+    if (this.dataSource instanceof Layer) {
+      this.dataSource.addSourceField(this.column);
```
VictorVelarde: I would suggest renaming dataSource to something else, like 'origin'. Everytime I read dataSource I just can't avoid thinking about just 'Source' and not about 'Layer'.

VictorVelarde (commented):
LGTM, but I would like to see some tests, to validate this new approach. I'm also not sure about the work on Histogram. Shouldn't that be in a different PR?

src/lib/viz/dataview/mode/DataViewMode.ts
manmorjim: renamed to `dataOrigin`

jesusbotella:
It looks good.

Actually, I was thinking that maybe it's not necessary to apply that `operation__columnName` column name pattern in the aggregation. That way data would be easier to parse and we wouldn't need to check if tile data comes aggregated, and filters would work.

I think I discarded it in first place because of collisions, but given that no original column is included when data is aggregated it would work much better than now. What do you think? @VictorVelarde @manmorjim

VictorVelarde:
Wouldn't we have, in any case, several aggregations at the same time?

jesusbotella:
Yeah, that's why I discarded it, nvm 😂. It seems like I need to write things down so that I don't forget.

VictorVelarde (approved):
🚀

PR #86, opened 2020-07-16
Use aggregated data for calculations in HistogramDataView

jesusbotella:
This PR implements usage of aggregated columns in MVT for DataView calculations.


src/lib/viz/dataview/histogram/HistogramDataViewImpl.ts
```diff
@@ -1,17 +1,19 @@
...
 import { DataViewRemote } from '../mode/DataViewRemote';
 import { DataViewOptions } from '../DataView';
+import { getFeatureValue } from '../utils';
 
 export class HistogramDataViewImpl extends DataViewImpl<HistogramDataViewData> {
   options: HistogramDataViewOptions;
 
   constructor(dataView: DataViewMode, options: HistogramDataViewOptions) {
-    super(dataView, { operation: AggregationType.COUNT });
+    super(dataView, { operation: AggregationType.AVG });
```
VictorVelarde, line 16: What's the reason behind this change? Shouldn't we count frequencies per bin and later on consider the clustered-count to increase its weight?

VictorVelarde: Maybe we can add some specs around this, to better grasp & validate its calculation method.

jesusbotella: The thing is that the operation does literally nothing. I had to put something there and I think I changed it for a strange thought that I cannot remember. So we can revert this change.

We do count frequencies by bin with `_cdb_feature_count` and sum them to the bin that they belong to, but I don't really understand your approach.

We don't have a weight property for histograms either, so we are just adding the features to the bin that they belong to without applying any common weight for all of them.

jesusbotella: Sorry, I was wrong on the operation part (as always 😂 )

The change is because the operation defines the aggregation for the column that contains the value of the bin. `count` didn't make sense, so I changed it to `avg`.

VictorVelarde (approved):
👍

PR #92, opened 2020-07-17
fix agg columns reinstantiation

simon-contreras-deel:
when local filtered and agg columns exist, we are always doing a reinstantiation with the very same values


manmorjim (approved):
nice! :rocket:

jesusbotella (approved):

PR #93, opened 2020-07-17
Fix AVG aggregated data

jesusbotella:
The examples I presented this morning included a fix for AVG and SUM calculation.

As we are getting already sum up data in SUM, there's no need to multiply it by feature count another time. In fact it was returning wrong data. But we need to do that in AVG, so I changed it a little bit :)


src/lib/data/operations/aggregation/feature/feature-aggregation.ts
```diff
@@ -36,22 +36,26 @@ const aggregationFunctions: Record<AggregationType, Function> = {
...
 
   [AggregationType.AVG](aggregatedFeatures: AggregatedFeatureProperties[]) {
-    return (
-      aggregationFunctions.sum(aggregatedFeatures) / aggregationFunctions.count(aggregatedFeatures)
-    );
+    let totalFeatureCount = 0;
+
+    const totalSum = aggregatedFeatures.reduce((total, feature) => {
+      totalFeatureCount += feature.featureCount;
+      return total + Number(feature.aggregatedValue) * feature.featureCount;
```
VictorVelarde, line 43: Ok, I see, so this is now the same ponderated logic as in VL. LGTM

VictorVelarde (approved):
🚀

src/lib/data/operations/aggregation/feature/feature-aggregation.ts
jesusbotella: Yes, it is. Using the other way, avg was really off, because of comparing "different" sets of values

PR #96, opened 2020-07-20
Allow filtering of Maps API aggregated data

jesusbotella:
This PR implements map filtering for data aggregated in tiles generated by Maps API.

As aggregated data has an specific column name with `_cdb_${operation}_{column}`, we didn't support filtering of those special columns until now.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 
 ## [Unreleased]
 
+### Added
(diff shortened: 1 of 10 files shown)


VictorVelarde (commented):
LGTM. I just miss some specs... And isn't any change required for Category dataview?

jesusbotella:
As we talked on Friday, it is not necessary to make changes in Category Widget because we filter by category name, not by category value.

VictorVelarde (approved):
Nice! Please add a brief note to the changelog, and we're ready to go

PR #97, opened 2020-07-22
dataview with remote source

simon-contreras-deel:
Fix: https://app.clubhouse.io/cartoteam/story/93762/dataview-can-not-be-instantiated-from-sources


src/lib/viz/dataview/DataView.ts
```diff
@@ -85,6 +85,14 @@ export abstract class DataView<T> extends WithEvents {
     return data;
   }
 
+  public on(type: string, handler: mitt.Handler) {
+    if (type === 'dataUpdate' && this.dataOrigin instanceof Source) {
```
VictorVelarde: don't we have any const that we can re-use broadly?

VictorVelarde (approved):
Nice, drop a line with a short fix description

VictorVelarde (approved):
Feel free to merge if you see it fine

PR #98, opened 2020-07-22
feat: Override default property for style helpers

padawannn:


src/lib/viz/style/helpers/color-bins-style.ts
```diff
@@ -21,6 +22,8 @@ export interface ColorBinsOptionsStyle extends Partial<BasicOptionsStyle> {
   nullColor: string;
   // Color applied to features which the attribute value is not in the breaks.
   othersColor: string;
+  // Styling property.
+  property?: ColorsProperties;
```
VictorVelarde: Nice!

src/lib/viz/style/helpers/color-categories-style.ts
```diff
@@ -19,6 +20,8 @@ export interface ColorCategoriesOptionsStyle extends Partial<BasicOptionsStyle>
   nullColor: string;
   // Color applied to features which the attribute value is not in the breaks.
   othersColor: string;
+  // Styling property.
+  property?: ColorsProperties;
```
VictorVelarde: Nit. I would probably go for singular naming here for the type, so `ColorProperty` (and the same with `SizeProperty` type). What do you think?

src/lib/viz/style/helpers/size-categories-style.ts
```diff
@@ -112,12 +115,16 @@ function calculateWithCategories(
...
-    };
+    if (!options.property || options.property === 'size') {
+      obj = {
+        getRadius: getSizeValue,
+        pointRadiusMinPixels: minSize,
+        pointRadiusMaxPixels: maxSize,
+        radiusUnits: 'pixels'
+      };
+    } else if (options.property === 'strokeWidth') {
+      obj = { getLineWidth: getSizeValue };
```
VictorVelarde: Should we set the line widht Units param to pixels too?

VictorVelarde (commented):
@padawannn Overall it looks nice. I just did a comment on naming. 

I think we should probably add some basic unit tests to validate 'correct' inputs and deckgl functions set on every case. And also a simple example in the examples/debug/styles folder for playing a bit with it (we could later decide if we could promote that example to the public folder)

src/lib/viz/style/helpers/color-categories-style.ts
padawannn: Done

src/lib/viz/style/helpers/size-categories-style.ts
padawannn: Done

padawannn:
The example is https://github.com/CartoDB/web-sdk/pull/98/files#diff-4b7e7f58a59ba200eb20aeb64093a158 and I've added some test for each helper

examples/_debug/styles/color-bins-style- override-property.html
```diff
@@ -0,0 +1,105 @@
...
+          strokeWidth: 4,
+          color: '#FFF',
+          property: 'strokeColor'
+        });
+        const layer = new carto.viz.Layer('listings_madrid', style);
+        layer.addTo(deckMap);
+      }
+
+      function initializeLines() {
+        carto.auth.setDefaultCredentials({ username: 'cartoframes' });
```
VictorVelarde, line 64: Just a comment. 👍 This is OK for debug. For public examples, we should always use 'public' user

VictorVelarde: By the way, the example looks really nice. 

It is not a priority right now, but we would probably want to document this new feature a bit better in the Styles guide, with a note like:

>>> With color helpers, the surface of the feature is coloured by default (thus, there is an implicit visual property: `color`, which we considered the most common one). But if you want to use the stroke, you can specify it, with something like for example:

```js
carto.viz.style.colorBins('availability_365', { strokeWidth: 4, property: 'strokeColor' });
```

image.

And the same could be applied for size.

cc/ @borja-munoz

VictorVelarde: Every step we take on this direction, we're near this 'advanced VL' example: https://carto.com/developers/carto-vl/examples/maps/advanced/taxi-pickups.html which is to me one of the most complex examples I remember: several properties at the same time (+ animation)

src/lib/viz/__tests__/styles/ColorBinsStyle.spec.ts
```diff
@@ -131,6 +131,31 @@ describe('ColorBinsStyle', () => {
         expect(error).toBeInstanceOf(CartoStylingError);
       }
     });
+
+    it('If geometryType is Polygon and property is strokeColor getLineColor should be a function', () => {
```
VictorVelarde, line 135: Really nice!

src/lib/viz/__tests__/styles/ColorCategoriesStyle.spec.ts
```diff
@@ -127,6 +127,36 @@ describe('ColorCategoriesStyle', () => {
         expect(error).toBeInstanceOf(CartoStylingError);
       }
     });
+
+    it('If geometryType is Polygon and property is strokeColor getLineColor should be a function', () => {
```
VictorVelarde, line 131: 👌

VictorVelarde (approved):
Really nice work. Add a line in the changelog like this, and we're ready to go 🚀 

###Added
- New option to specify the visual property to use in color ('color' | 'strokeColor') and size ('size' | 'strokeWidth') style helpers ([#98](https://github.com/CartoDB/web-sdk/pull/98))

padawannn:
@VictorVelarde changelog updated

PR #99, opened 2020-07-23
fix DATA_READY_EVENT

simon-contreras-deel:
After some changes in the first event PR, I did not realized DATA_READY_EVENT was not fired. This PR is fixing that


VictorVelarde (approved):
👍

PR #100, opened 2020-07-23
[Legend] basic

padawannn:
It's a first PR to check if my approach is ok.
Some considerations:
- This branch comes from feature/ch94386/style-override-default-property-for-style branch.
- The last Legend parameter is options and it can be used for overwriting legend properties allowing you to customize the legend. In the example, I'm overwriting the label property.
- In the Style.ts _legendProperties will be optional while we develop the other style helpers.
- There are two TODOs that we need to discuss in the daily.


padawannn:
@VictorVelarde I've merged this branch with padawannn/layer-is-ready so this PR is ready to merge. The options LegendWidget param is a first approach but we need to discuss about this to maintain CF compatibility

VictorVelarde:
I'm having a look at this one. One question... should it work for layers with **default style** (no style helper assigned)?

For example, this is use case...
image

EDIT:

@padawannn after having a quick look, it seems that evalFNLegend in `basic-style.ts` is not getting the 'options' and it could

src/lib/viz/legend/LegendWidget.ts
```diff
@@ -0,0 +1,57 @@
...
+export class LegendWidget {
+  constructor(element: string | HTMLElement, layer: Layer, options = {}) {
+    const domElement = queryDOMElement(element);
+    validateParameters(domElement, layer, options);
+    const legendWidget = domElement as any;
+
+    if (layer.isReady()) {
+      legendWidget.data = layer.getLegendData(options);
+    } else {
+      layer.on('dataReady', () => {
```
VictorVelarde: Let's try to use internally our enums for event names, in this case `LayerEvent.DATA_READY`, currently inside of Layer.ts

examples/widget/categoryWidget.html
```diff
@@ -117,7 +117,7 @@ <h4 class="as-subheader as-mb--12">
         });
 
         /*
-          There are two option availables: 
+          There are two option availables:
```
VictorVelarde: typo --> "two available options"

src/lib/viz/legend/LegendWidget.ts
```diff
@@ -0,0 +1,57 @@
...
+import { CartoError } from '@/core/errors/CartoError';
+
+export type LegendGeometryType = 'point' | 'line' | 'polygon';
+
+export type LegendProperties = {
+  type: LegendGeometryType;
+  color?: string;
+  marker?: string;
+  strokeColor?: string;
+  strokeStyle?: string;
```
VictorVelarde, line 12: Sorry, what's strokeStyle?

src/lib/viz/legend/LegendWidget.ts
```diff
@@ -0,0 +1,57 @@
...
+  label?: string;
+  width?: number;
```
VictorVelarde, line 14: Out of curiosity: Why are we using here `width` instead of `size` (as in the style)?

For further discussion: 
- should we include also other style properties like `opacity`? 
- I'm also seeing that we have a block of 'layer-style-related' props and 'others' (more specific to the Legend itself). Maybe we should try to re-organize them in 2 levels, so `LegendProperty` type could have internally a `Style` property for overrides... what do you think?

src/lib/viz/legend/LegendWidget.ts
```diff
@@ -0,0 +1,57 @@
+import { Layer } from '@/viz/layer';
+import { queryDOMElement } from '@/core/utils/dom';
+import { CartoError } from '@/core/errors/CartoError';
+
+export type LegendGeometryType = 'point' | 'line' | 'polygon';
+
+export type LegendProperties = {
```
VictorVelarde, line 7: ~I think we should also use here the singular for the type, thus **LegendProperty**~

src/lib/viz/style/Style.ts
```diff
@@ -1,18 +1,27 @@
 import { GeoJsonLayerProps } from '@deck.gl/layers/geojson-layer/geojson-layer';
+import { LegendProperties } from '@/viz/legend';
 import { CartoStylingError, stylingErrorTypes } from '../errors/styling-error';
 import { StyledLayer } from './layer-style';
 
 export type StyleProperties =
   | GeoJsonLayerProps<any>
   | ((layerStyle: StyledLayer) => GeoJsonLayerProps<any>);
 
+type LegendPropertiesFunction = (layerStyle: StyledLayer, options: any) => LegendProperties[];
```
VictorVelarde, line 10: IMO after the rename one part could stay in plural and the type in singular, like:
```
type LegendPropertiesFunction = (layerStyle: StyledLayer, options: any) => LegendProperty[];
```

What do you think?

src/lib/viz/style/helpers/basic-style.ts
```diff
@@ -7,5 +8,26 @@ export function basicStyle(options: Partial<BasicOptionsStyle> = {}) {
...
+    if (!meta.geometryType) {
+      return [];
+    }
+
+    return [
+      {
+        type: meta.geometryType.toLocaleLowerCase() as LegendGeometryType,
+        color: options.color,
+        strokeColor: options.strokeColor,
+        width: options.size,
```
VictorVelarde: As I said before, I couldn't get these 'options' working if using basicStyle

VictorVelarde (commented):
This is a nice approach. I think we should fix the basicStyle use case and that's it.

IMO we should add also at least some basic specs.

I suggested also some refactors, but you're free to consider them outside of this PR, if you want to push it faster.

src/lib/viz/legend/LegendWidget.ts
VictorVelarde: My mistake forget it, LegendProperties is the right choice!!

padawannn:
> I'm having a look at this one. One question... should it work for layers with **default style** (no style helper assigned)?
> 
> For example, this is use case...
> image
> 
> EDIT:
> 
> @padawannn after having a quick look, it seems that evalFNLegend in `basic-style.ts` is not getting the 'options' and it could

You're right. It is already fixed

src/lib/viz/legend/LegendWidget.ts
padawannn: Done!

examples/widget/categoryWidget.html
padawannn: Done!

src/lib/viz/legend/LegendWidget.ts
padawannn: A bit explanation about LegendProperties:
LegendProperties are the list of the properties that are supported by airship legend widgets. You can see these properties here https://carto.com/developers/airship/reference/#/components/legends?a=props
As you can see there is a strokeStyle property with these possible values https://developer.mozilla.org/en-US/docs/Web/CSS/border-style
Our styles helpers don't support this property but I think that can be interested If the user want to set it manually but we could remove it.

src/lib/viz/legend/LegendWidget.ts
padawannn: As I explain in the previous comment, LegendProperties are the list of the properties that are supported by airship legend widgets so this property should be `width` because `size` is not supported by airship legend widgets.

About opacity, this property doesn't exist in airship legend widgets but checking CartoFrames I have seen that they include the opacity like the fourth component of the RGBA color so I've changed this PR to include the opacity and I'm going to change in the others legends PRs.

I agree with you to re-organize `LegendProperty` in two levels but I would like to do a bit of research and do this in other PR that will affect all legends PRs

src/lib/viz/style/helpers/basic-style.ts
padawannn: Fixed, check my last comment

src/lib/viz/legend/LegendWidget.ts
VictorVelarde: Yeah, we should probably review them. We don't need to repeat all the previous props if we think that some don't make sense now.. Let's postpone the decission on this. Thx!

VictorVelarde (approved):
🚀 Let's merge!

VictorVelarde:
Just update with develop and let's move on

PR #102, opened 2020-07-24
[Legend] colorBins

padawannn:
In this case, I think that point legend and line legend has no sense 
Captura de pantalla 2020-07-24 a las 14 44 53

Captura de pantalla 2020-07-24 a las 14 45 33

And we should always use the polygon type legend:

Captura de pantalla 2020-07-24 a las 14 45 12

In fact, Airship documentation doesn't have as-legend-color-bins-point and as-legend-color-bins-line example. If the user wanted to force one of these two legends they could use the property params but by default, I would set polygon like type in the color-bins helper (in this PR I'm not forcing the polygon type)

What do you think?


VictorVelarde:
What happens in CF? Isn't the type adjusted to the geometry type of the Layer?
I think we should probably do the same. I tend to think just "polygons" for all options "feels wrong", but not 100% sure...

padawannn:
mmm, I'm going to compare each case with Carto frames because they use intervals for each legend value for color-bins legend and it can work for all geometries types.
https://carto.com/developers/cartoframes/examples/#example-color-bins-style

src/lib/viz/style/helpers/color-bins-style.ts
```diff
@@ -62,17 +63,57 @@ export function colorBinsStyle(
...
+        ...properties
+      };
+    });
   };
 
-  return new Style(evalFN, featureProperty);
+  return new Style(evalFN, featureProperty, evalFNLegend);
+}
+
+function getBreaks(opts: ColorBinsOptionsStyle, meta: SourceMetadata, featureProperty: string) {
```
VictorVelarde, line 108: A question related to this. I see you have extracted the `getBreaks`function from the original place, which is nice. But I also see a small change regarding to how the number of breaks is passed. Is it keeping the same behaviour (part of the refactor) or is related with a detected bug on this part?

src/lib/viz/style/helpers/color-bins-style.ts
```diff
@@ -62,17 +63,57 @@ export function colorBinsStyle(
...
+      return [];
     }
 
-    return calculateWithBreaks(featureProperty, opts.breaks, meta.geometryType, opts);
+    const opts = defaultOptions(meta.geometryType, options);
+    const breaks = getBreaks(opts, meta, featureProperty);
+    const stats = meta.stats.find(f => f.name === featureProperty) as NumericFieldStats;
+    let ranges = [...breaks, stats.max];
+    const colors = getColors(opts.palette, ranges.length);
+    ranges = [stats.min, ...ranges];
```
VictorVelarde, line 86: Well done ✨ 
At the beginning I have doubts, about them being 'open' <X or >X (first and last intervals), but it's nice to have here the real min and max values, so the interval is 'real'.

VictorVelarde (approved):
LGTM, nice PR!

I just miss some tests, but if we want to unlock the integration we can merge this in first, and add the specs once you're back

VictorVelarde:
Let's wait just a bit for PR #100, and then you can merge this one!

src/lib/viz/style/helpers/color-bins-style.ts
padawannn: If I have not been wrong, it is keeping the same behavior. In the original code in the line 69 you can see `return calculateWithBreaks(featureProperty, breaks, meta.geometryType, opts);` and in the line 72 `return calculateWithBreaks(featureProperty, opts.breaks, meta.geometryType, opts);` The only difference between both calls is the `breaks` or `opts.breaks` parameter. In the PR getBreak function returns `breaks` or `opts.breaks`

PR #103, opened 2020-07-27
Fixed some bugs in addInTheRightPosition function

sclebal:
The function to add the layer in the right position (addInTheRightPosition) wasn't working properly.


VictorVelarde:
Just for the record, at the beginning a flip was made on the meaning of 'before/after', but we then decided to go after a rename, cause  ‘before’ and ‘after’ weere not very clear terms, using ‘overLayerId’ & ‘underLayerId’ as a better one now

VictorVelarde (approved):

PR #104, opened 2020-07-28
[Legend] colorCategories

padawannn:


src/lib/viz/style/helpers/color-categories-style.ts
```diff
@@ -53,27 +54,69 @@ export function colorCategoriesStyle(
...
     validateParameters(opts);
 
-    let categories;
+    const categories = getCategories(opts, meta, featureProperty).slice(0, opts.top);
 
-    if (opts.categories.length) {
-      categories = opts.categories;
-    } else {
-      const stats = meta.stats.find(c => c.name === featureProperty) as CategoryFieldStats;
+    return calculateWithCategories(featureProperty, categories, meta.geometryType, opts);
```
VictorVelarde, line 59: I'm gonna propose a refactor for a second phase on style helpers (not now).

Now that we have also 'legends' code inside the helpers we might need to rename certain functions... We might even.consider falling back into classes + factory methods, instead of keeping the pure functions... we'll see 😄

examples/_debug/legends/color-categories-style.html
```diff
@@ -0,0 +1,132 @@
+<!DOCTYPE html>
```
VictorVelarde, line 1: Please rename the example (and all of the new ones inside 'legends' to the camelCase way, so 'colorCategories' in this case.

In fact, the style helpers are the only filenames we left (by mistake or lack of time) with the old 'snake-case-naming'. We'll plan a small refactor on that, once we have the legends.

VictorVelarde (approved):
🚀 
Pending tests for a next PR

examples/_debug/legends/color-categories-style.html
padawannn: What a bad luck I am because I have copied all the examples from the styles folder XD

PR #108, opened 2020-07-29
fix: replaceDeckGLLayer not considering latest updated layers

VictorVelarde:
Ensures the latest layers state is considered when replacing a deckgl layer (it could have changed while awaiting for its creation)


sclebal (approved):

sclebal:
👌 LGTM @VictorVelarde

PR #109, opened 2020-07-29
[Legend] sizeContinuous

padawannn:


examples/_debug/legends/size-continuous-style.html
```diff
@@ -0,0 +1,102 @@
...
+          // bins: 3,
+          // sizeRange: [1, 5]
+        });
+        const layer = new carto.viz.Layer('roads', style);
+        layer.addTo(deckMap);
+        new carto.viz.Legend('#legend', layer);
+      }
+
+      initialize();
+      // initializeLines();
```
VictorVelarde, line 99: I've noticed that this legend, for lines, get too wide:

image

Not an issue here, but on the Airship side...

VictorVelarde (approved):
🚀

PR #112, opened 2020-07-30
Fix wrong path in modules & lib folder refactor

VictorVelarde:
main & module properties in package.json were not pointing to the correct path


VictorVelarde:
I have also some doubts also about aliased imports (if that could be causing some issues on cjs too)

Right now, using npm alpha.1

image

CHANGELOG.md
```diff
@@ -10,13 +10,15 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 - Aggregated data filtering in widgets ([#96](https://github.com/CartoDB/web-sdk/pull/96))
 
 ### Changed
-- Rename `beforeLayerId|afterLayerId` options in `carto.viz.Layer.addTo` to a more clear `overLayerId|underLayerId`. 
+
```
jesusbotella, line 13: This blank line is not needed

package.json
```diff
@@ -12,8 +12,8 @@
   "files": [
     "dist"
   ],
-  "main": "dist/cjs/index.js",
-  "module": "dist/esm/index.js",
+  "main": "dist/cjs/lib/index.js",
```
jesusbotella: This solves the problem, but what we should really need to do is to remove lib folder. We don't have any other code apart from the library one, and `src/` folder doesn't have specific files.

jesusbotella (approved):
Looks good, but as I said in the comment, I think we should remove `lib/` folder so that child folders are at `src/` level.

jesusbotella:
About the aliased imports, it is true that those don't work out of the box. We'd need to use a Babel plugin like this: https://www.npmjs.com/package/babel-plugin-module-resolver

package.json
VictorVelarde: done!

PR #113, opened 2020-07-31
feat: legend options

padawannn:
The legend options that are working in this PR:
- samples: for continuous helpers (in other helpers have no sense),
- order: for all helpers except categories helpers (I need to check better how it's working in CF).
- othersLabel: for categories helpers (in other helpers have no sense).

format it's accepted but not working yet. I need to check CF because the VL Bridget works with strings, numbers, and arrays like params for format function.

dynamic is accepted but not working yet.


VictorVelarde:
Label for basicStyle / default style was not working on the legend, so I'm adding that in some commits, to allow getting these "entry legend texts" 
image

What's displayed is a new explicit `label`parameter, if set for the legend (or just the 'layerId' if it is not specified)

VictorVelarde (approved):
After fixing basic legend, everything seems to be working, so let's move on. I'm gonna create another story for other possible 'extra params' (horizontal / vertical...)

PR #115, opened 2020-07-31
fix: add version variable to esm & cjs modules

VictorVelarde:
CJS & ESM modules didn't have access to WEBSDK_VERSION (like umd with webpack does)


manmorjim (approved):
LGTM :+1:

PR #116, opened 2020-08-05
[LEGENDS] Create unit tests

manmorjim:
Some fixes added to this PR:
- Make opacity compatible with RGBAColor format
- Fix max value label for continuous style
- Fix order parameter validation
- Add order for `sizeContinuousStyle`


josema-xyz (approved):
Checking conditions! :muscle:

PR #117, opened 2020-08-10
Use the feature geom instead of the pixel clicked

manmorjim:
Use the feature geometry when it is a point in order to improve the precision of the popup position


VictorVelarde:
I have doubts about the implementation, so not merging yet... 

The (required) mixed used of SQL api and Maps API seem to be too complex for the feature. I would like to think about it a bit more...

PR #118, opened 2020-08-11
FIX shared columns for multiple widgets

manmorjim:


josema-xyz (approved):
Working now! Thank you @manmorjim!

PR #120, opened 2020-08-12
[LEGENDS] Dynamic legends

manmorjim:


josema-xyz (approved):
Most changes converting calls from sync to async. Async hell... well, it's the browser life :shrug: 

:rocket:

PR #125, opened 2020-08-25
Allow custom id property for Source + id-related fixes

VictorVelarde:
Fixes
- Fix interactivity styles on GeoJSON, where no cartodb_id is available

Features (internal, not currently exposed)
- Allow specifying a custom id property for a MVTLayer (in viewportFeaturesGenerator), being that a number but also now a string. That's not used right now for SQLSource, as there it is a hardcoded 'cartodb_id', but it is required if we consider another arbitrary tile sources (eg BQ), as it is the key to gather different pieces in tiles with good viewport calculations

Refactor
- I did also some small refactors, doc headers and debug examples renames


examples/_debug/source/geojson/noCartoDbId.html
```diff
@@ -0,0 +1,94 @@
...
+        // style
+        const style = carto.viz.style.colorBins('value', {
+          size: 30,
+          palette: 'ORYEL',
+          viewport: true
+        });
+
+        // geojson source (external, to allow custom options)
+        const geojsonSource = new carto.viz.source.GeoJSON(data);
+        // const geojsonSource = new carto.viz.source.GeoJSON(data, { uniqueIdProperty: 'name' });
```
padawannn, line 66: For this example, should you uncomment line 66 and comment line 65?

src/viz/interactivity/viewport-features/ViewportFeaturesGenerator.ts
```diff
@@ -58,43 +152,18 @@ export class ViewportFeaturesGenerator {
...
-        return selectPropertiesFrom(feature.properties as Record<string, unknown>, properties);
-      })
-      .flat();
+    return this._deckLayer.state.tileset.selectedTiles;
   }
 
   private getMVTViewportFilteredFeatures(selectedTiles: ViewportTile[], properties: string[]) {
     const currentFrustumPlanes = this.getViewport().getFrustumPlanes();
-    const featureCache = new Set<number>();
+    const featureCache = new Set<string>(); // don't assume just number (the most common use case, eg cartodb_id)
```
padawannn, line 166: Could be better new Set<string|number>() ? what do you think?

src/viz/interactivity/viewport-features/ViewportFeaturesGenerator.ts
```diff
@@ -146,51 +215,88 @@ export class ViewportFeaturesGenerator {
     return isInside;
   }
 
-  private getSelectedTiles(): ViewportTile[] {
-    if (!this.deckLayer || !this.deckLayer.state) {
-      return [];
+  private getFeatureId(feature: GeoJSON.Feature): string {
+    if (!feature.properties) {
```
padawannn, line 219: We cannot get a feature without properties but with feature id?

src/viz/interactivity/viewport-features/ViewportFeaturesGenerator.ts
```diff
@@ -146,51 +215,88 @@ export class ViewportFeaturesGenerator {
...
 }
 
+// #region Private complements
 interface ViewportFeaturesGeneratorOptions {
   uniqueIdProperty?: string;
 }
 
 interface InsideViewportCheckOptions {
-  featureCache: Set<number>;
+  featureCache: Set<string>;
```
padawannn, line 297: Could be better Set<string|number>?

padawannn (changes requested):

examples/_debug/source/geojson/noCartoDbId.html
VictorVelarde: Well, it's a different use case... The main fix is the first one, using a geojson file without 'cartodb_id' (that was working wrong on interactivity). Adding a uniqueProperty to GeoJSON is indeed an option I'm now considering if it makes sense to expose or not.

src/viz/interactivity/viewport-features/ViewportFeaturesGenerator.ts
VictorVelarde: That's a good point. I had assumed it "contained" in the properties, but I guess we should also consider the feature.id option

src/viz/interactivity/viewport-features/ViewportFeaturesGenerator.ts
VictorVelarde: That's the idea, but with an explicit conversion from number (integer expected) always to string. Let me have a look at it, to see if I can refactor it

VictorVelarde:
Closing in favour off https://github.com/CartoDB/web-sdk/pull/128 (with updated lock file)

PR #126, opened 2020-08-25
Fix building process with ESM and CommonJS

alasarr:
This PR fixed the building process with ESM and CommonJS.

The problem is that babel removes the interfaces from the modules and it fails when it tries to import.

```
./node_modules/@carto/web-sdk/dist/esm/viz/source/index.js
Attempted import error: 'AggregatedColumn' is not exported from './Source'.
```

The solution, according to [this thread]( https://github.com/babel/babel/issues/8361) is to specify it's a type when export interfaces.


VictorVelarde (approved):
🚀

PR #129, opened 2020-08-26
Enable library for frameworks via NPM modules

alasarr:
This PR brings support to use the library via npm.

With this branch, I've been able to include the SDK in a react project (I know I'm not using the DeckGL react component, just want to try if I can ingest the library via npm without importing the umd by hand). 

It should work on Angular or Vue, but we should try it.

```javascript
import React from 'react';
import { auth, viz } from '@carto/web-sdk';

class Map extends React.Component {
  map = null

  componentDidMount() {
    auth.setDefaultCredentials({ username: 'public' });
    this.map = viz.createMap();
    const ports = new viz.Layer('world_ports');
    ports.addTo(this.map);
  }

  render() {
    return (
      <div id='map'>
        Map
      </div>
    )
  }
}

export default Map;
```

It also removes the dependency of DeckGL because it's only available via [Scripting API](https://deck.gl/docs/get-started/using-standalone#using-the-scripting-api) and it includes a dependency for Mapbox GL, so we don't need to add the link to mapbox CDN.

TODO: remove all the links to Mapbox at the examples.


VictorVelarde:
I have just updated the typings after your contribution

PR #130, opened 2020-08-27
Feature/ch100368/performance getfeatures

padawannn:
Performance:
- Debounce function for DATA_CHANGED layer event
- Prevent replace layer every time widget data changes
- Move debounce function to root utils file


src/viz/dataview/DataView.ts
```diff
@@ -3,9 +3,10 @@ import { Layer, Source } from '@/viz';
 import { Credentials } from '@/auth';
 import { Filter, ColumnFilters, SpatialFilters } from '@/viz/filters/types';
 import { AggregationType } from '@/data/operations/aggregation/';
+import { debounce } from '@/viz/utils';
```
VictorVelarde: how about moving it to **core/utils**? Right now the 'utils' are a bit of a mess, but IMO we should start organizing them better, into different categories.

src/viz/layer/Layer.ts
```diff
@@ -457,13 +464,18 @@ export class Layer extends WithEvents implements StyledLayer {
     dimensions.forEach(dimension => this._source.addField(dimension));
     columns.forEach(aggregatedColumn => this._source.addAggregatedColumn(aggregatedColumn));
 
-    return this.replaceDeckLayer();
+    if (this._source.needsInitialization) {
```
VictorVelarde, line 467: Nice!

src/viz/layer/Layer.ts
```diff
@@ -688,19 +700,19 @@ export class Layer extends WithEvents implements StyledLayer {
    * Manage data state and asociated events
    */
   private _sendDataEvent(referer: 'onViewportLoad' | 'onAfterRender') {
-    const isGeoJsonLayer = this._source.sourceType === 'GeoJSON';
+    const isGeoJsonLayer = this._source.sourceType === 'GeoJSONSource';
```
VictorVelarde: After my last merge, this is 'GeoJSON' not 'GeoJSONSource' (maybe it's not the best move, but I did it and it won't work otherwise...)

src/viz/dataview/DataView.ts
padawannn: Done

src/viz/layer/Layer.ts
padawannn: Done

src/viz/dataview/DataView.ts
```diff
@@ -3,9 +3,10 @@ import { Layer, Source } from '@/viz';
 import { Credentials } from '@/auth';
 import { Filter, ColumnFilters, SpatialFilters } from '@/viz/filters/types';
 import { AggregationType } from '@/data/operations/aggregation/';
+import { debounce } from '@/viz/core/utils';
```
VictorVelarde, line 6: Really nice!

VictorVelarde (approved):
🚀 Well done!

Please add a note to changelog, like this and merge when you want:

### Added
- Better performance in `Layer`, with a debounced approach to DATA_CHANGED event ([#130](https://github.com/CartoDB/web-sdk/pull/130/))

PR #131, opened 2020-08-28
feat:view port features queue

padawannn:


src/viz/interactivity/viewport-features/ViewportFeaturesQueue.ts
```diff
@@ -0,0 +1,30 @@
+export class ViewportFeaturesQueue {
```
VictorVelarde, line 1: This looks like an ingenious and practical approach. Cool! ✨

src/viz/layer/Layer.ts
```diff
@@ -510,14 +514,23 @@ export class Layer extends WithEvents implements StyledLayer {
...
+        if (this.filtersCollection.hasFilters()) {
+          features = features.filter(feature => filters.applicator(feature));
+        }
+
+        this._viewportFeaturesQueue.resolveQueue(features);
+      });
     }
 
-    return features;
+    return (await pendingPromise.catch(error => console.debug(error))) || [];
```
VictorVelarde, line 533: I'm wondering if we could hide a bit the complexity from the queue, under the ViewportFeaturesGenerator itself....

I'm thinking about instantiating _viewportFeaturesQueue inside it, and let the public method getFeatures make the IF-EMPTY-XXXX internally.... what do you think?

VictorVelarde (approved):
Great! Just a general comment.

VictorVelarde:
Let's wait just to have first #130  and we're ready to go

src/viz/layer/Layer.ts
padawannn: I like this approach but we would need to refactor some things. I prefer to keep it as it is for the moment

PR #136, opened 2020-09-01
Fix: set deckInstance from the beginning when using addTo

VictorVelarde:
To avoid some errors with DataViewRemote


sclebal (approved):
LGTM 👍

PR #137, opened 2020-09-01
Feat: Add web-sdk version to Maps api client requests

VictorVelarde:
- add **client**, as in VL: a param like this is added: `https://public.carto.com/api/v1/map?api_key=default_public&client=websdk-1.0.0-alpha.2`
- refactor maps/ folder


VictorVelarde:
Hi, @dgaubert! Just pinging you to ensure 'client' param is still fine (exactly the same approach in carto-vl client)

dgaubert:
Yes, it is! It should work in the same way as in Carto-VL.

padawannn (approved):

PR #138, opened 2020-09-01
Remove unneeded mapviews reinstantiations (remote dataviews performance)

padawannn:
- Remove TILES_LOADED event (use DATA_CHANGED instead). I keep this event commented in case we use it in the future.
- I've tried different approaches and with the current code, the easier way is to have a cache in MapsDataviews.ts. Another approach requires a big refactor

### DIFF
--- a/src/maps/MapsDataviews.ts
+++ b/src/maps/MapsDataviews.ts
@@ -6,10 +6,12 @@ import { Client, MapDataviewsOptions } from './Client';
 export class MapsDataviews {
   private _source: string;
   private _mapClient: Client;
+  private _layergroupCache: Record<string, any>;
 
   constructor(source: string, credentials: Credentials) {
     this._source = source;
     this._mapClient = new Client(credentials);
+    this._layergroupCache = {};
   }
 
   public async aggregation(params: Partial<MapDataviewsOptions>): Promise<AggregationResponse> {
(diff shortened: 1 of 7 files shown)


src/maps/MapsDataviews.ts
```diff
@@ -55,18 +57,27 @@ export class MapsDataviews {
...
     dataviewName: string,
     type: string,
     options: Partial<MapDataviewsOptions>
   ) {
+    const layergroupCacheKey = `type_${JSON.stringify(options)}`;
+
+    if (this._layergroupCache[layergroupCacheKey]) {
+      return this._layergroupCache[layergroupCacheKey];
+    }
+
```
VictorVelarde, line 80: I see the trick, nice cache

src/viz/layer/Layer.ts
```diff
@@ -29,7 +29,7 @@ import { basicStyle } from '../style/helpers/basic-style';
 export enum LayerEvent {
   DATA_READY = 'dataReady',
   DATA_CHANGED = 'dataChanged',
-  TILES_LOADED = 'tilesLoaded',
+  // TILES_LOADED = 'tilesLoaded',
```
VictorVelarde: So eventually no use at all of this...? 

I guess that if we need to 'refresh' the tiles without changing the viewport (for example when requiring new fields in the map), we are rising an equivalent data changed event right?

If that's it, we should completely remove TILES_LOADED event

VictorVelarde (approved):
LGTM. I would remove TILES_LOADED at all, if we check it's not required anymore.

You can push this one first, and then I will try to apply the maps refactor from https://github.com/CartoDB/web-sdk/pull/137

src/viz/layer/Layer.ts
padawannn: ok, I've removed the comments

PR #139, opened 2020-09-02
Fix carto.viz.dataview.Formula not getting an updated source

padawannn:


VictorVelarde (approved):
Just add a line to the changelog describing the fix, please

PR #143, opened 2020-09-09
Add deprecation warning

VictorVelarde:


VictorVelarde:
Sorry, I'm moving on, as we're in a bit of a hurry

